Issue #17, opened 2018-07-20
Is it possible to set the character set to UTF-8 or a locale (type "es-ES" for example) "?


patryn32:
I just started doing tests with pdf-lib, and I'm wondering if you can set the character set to my locale (in this case 'es-ES') to be able to see all the characters correctly.
My program fills in a pdf form, and it gets it without problems. Although with the exception of course of the Spanish characters.
Here is my code.
[informevnpdflib.zip](https://github.com/Hopding/pdf-lib/files/2214154/informevnpdflib.zip)

Hopding:
Hello @patryn32. I'm afraid there is not currently a way to change the character set or locale. This is something that I'd definitely like to support, though. I've placed it on the [`TODO.md`](https://github.com/Hopding/pdf-lib/blob/master/TODO.md#features) file. 

I can't make any promises as to when I'll get around to implementing this. However, if non-english text is an absolute requirement for you, one (hacky) way to get it done would be to produce an image of the text (perhaps using [`node-canvas`](https://github.com/Automattic/node-canvas#canvastobuffer)) and draw that image onto the PDF. You can find an example of drawing an image on a PDF in [this example](https://github.com/Hopding/pdf-lib/tree/master/examples/document_modification/node/).

jerp:
Hi,

I can suggest to add Encoding 'WinAnsiEncoding' to the Font dictionary.

See https://github.com/jerp/pdf-lib/commit/199ecf2d67fa734f31f4caa332abe1a36460d8d1

Check out result in attached document (see "distribución") - I din't have the original doc, so I used the one in test-pdf directory

Jerome.
[InformeVN-user-20180828164746141.pdf](https://github.com/Hopding/pdf-lib/files/2328564/InformeVN-user-20180828164746141.pdf)

Hopding:
Version [`0.4.0`](https://www.npmjs.com/package/pdf-lib/v/0.4.0) is now published. It contains support for the WinAnsi character set when using one of the Standard 14 Fonts ([example](https://github.com/Hopding/pdf-lib/blob/master/examples/document_creation/node/index.js#L90)). The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v0.4.0).

You can install this new version with npm:
```
npm install pdf-lib@0.4.0
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@0.4.0/dist/pdf-lib.js
[...]

Issue #14, opened 2018-07-07
pdf-lib.min.js filesize is 1.1 MB, maybe is it possible to introduce a custom builds?


vitaly-zdanevich:
Not all users need all options, maybe from this library the most of us only need the ability of adding the texts.

Builder example http://fabricjs.com/build/:
![image](https://user-images.githubusercontent.com/3514015/42410752-cfbf67d2-81f7-11e8-8ab7-bbb8a43b864d.png)

1.1 MB is very big.

Hopding:
Yes, 1.1 MB is big for sure. The primary cause of this is the `fontkit` fork dependency. It's [`bundle.js`](https://github.com/Hopding/fontkit/blob/master/bundle.js) file is 895 KB. Reducing `pdf-lib`'s bundle size is currently an item on the [todo list](https://github.com/Hopding/pdf-lib/blob/master/TODO.md#dependencies).

I think the size of the fontkit `bundle.js` file could be reduced by optimizing the build process used for it. But I'm not sure to what extent. The official (non-forked) [build of fontkit](https://unpkg.com/fontkit@1.7.7/index.js) is still 519 KB. After minifying it with https://jscompress.com/, its size drops to 240 KB. This build doesn't work in all JS environments either, so it's unlikely that the forked build can become quite this small. But there's currently a 660 KB difference between the two. I'm reasonably sure that difference could be made significantly smaller.

Let's say the size of the forked build of fontkit could be reduced to 350 KB (total guess). The size of `pdf-lib.min.js` would then be 555 KB. So that's about a 50% reduction in its current size.

It may be, as you suggest, possible to build custom bundles that omit certain features of the library. In particular, it would be possible to create a build of `pdf-lib` that doesn't support embedding custom fonts (you would be restricted to using one of the [Standard 14 Fonts](https://github.com/Hopding/pdf-lib/blob/master/docs/modules/_core_pdf_document_standard14fonts_.md)). This build would not include the fontkit dependency at all, so it's bundle size would be about 205 KB. That's a size reduction of 81.4%.

I'm curious, are you intending to use the font embedding feature of `pdf-lib`?

vitaly-zdanevich:
Now I use `embedStandardFont('Helvetica')` according to the documentation.

Hopding:
[...]

Issue #13, opened 2018-07-07
Please adapt for Content Security Policy: drop Function(string), if possible (eval)


vitaly-zdanevich:
For example file
https://unpkg.com/pdf-lib/dist/pdf-lib.js
Line 30:
`var root = _freeGlobal || freeSelf || Function('return this')();`

Because of `Function('return this')` Chrome with CSP cannot execute this lib:
`Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed`

More information about CSP here https://developer.chrome.com/extensions/contentSecurityPolicy

Hopding:
I'm open to researching this issue. I think that particular line would have been inserted by [Rollup](https://rollupjs.org/guide/en) when building the `pdf-lib.js` and `pdf-lib.min.js` bundles, so it's probably not going to be fixed by a trivial change to `pdf-lib`'s source.

Can you please provide a minimum working example of the issue? An HTML file or zipped up collection of files should suffice.

Hopding:
I believe I've found the cause of the issue. I've created a minimal example that reproduces it:

[rollup-test.zip](https://github.com/Hopding/pdf-lib/files/2195741/rollup-test.zip)

You can unzip this example and run `npm install && npm run-script build` to create the `bundle.js` file with `rollup`. You'll see that the bundle looks like the following:
```javascript
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
	typeof define === 'function' && define.amd ? define(['exports'], factory) :
	(factory((global.Foo = {})));
}(this, (function (exports) { 'use strict';

	var commonjsGlobal = typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : {};

	/** Detect free variable `global` from Node.js. */
	var freeGlobal = typeof commonjsGlobal == 'object' && commonjsGlobal && commonjsGlobal.Object === Object && commonjsGlobal;

	var _freeGlobal = freeGlobal;

	/** Detect free variable `self`. */
	var freeSelf = typeof self == 'object' && self && self.Object === Object && self;

	/** Used as a reference to the global object. */
	var root = _freeGlobal || freeSelf || Function('return this')();
```
The cause of this seems to be importing a function directly from lodash, e.g. in the `foo.js` file of this example:
```javascript
// This type of import from lodash seems to be the cause:
import isString from 'lodash/isString';

[...]

PR #28, opened 2018-08-28
Add Encoding to Font Dictionary

jerp:
See https://github.com/Hopding/pdf-lib/issues/17


Hopding:
Hello @jerp. Thank you for submitting this PR! I've done a little bit of research on this and it looks like you're on the right track. However, I think a few more things will need to be done before this is complete.

`pdf-lib` currently writes text using Literal Strings. However, this means only ASCII characters can be used. This is obviously a severe restriction. What `pdf-lib` should do is convert strings of text into their character codes, and write them as Hex Strings. This is necessary to fully comply with the `WinAnsiEncoding`.

For example, this is how the text `Hello World!` would look as a Literal String:
```
(Hello World!)
```
And this is how it would look as a Hex String:
```
<48656C6C6F20576F726C6421>
```
Notice that each pair of numbers in the hex string is the hexadecimal representation of the corresponding character in the Literal String:
```
'Hello World!'
  .split('')
  .map(x => x.charCodeAt(0))
  .map(c => c.toString(16))
> [ '48', '65', '6c', '6c', '6f', '20', '57', '6f', '72', '6c', '64', '21' ]
```

However, JavaScript's `charCodeAt` method does not return `WinAnsiEncoding` character codes. It returns UTF-16 code points. Therefore, some conversion is necessary. This [webpage](http://string-functions.com/encodingtable.aspx?encoding=1200&decoding=1252) shows a comparison of the two encodings. `pdfkit` also has a nice mapping object that could be reused in `pdf-lib` as well: https://github.com/foliojs/pdfkit/blob/master/lib/font/afm.coffee#L58-L85.

-----

Please let me know if you're interested in working on these additional changes. I'm happy to help point you in the right direction and find your way around the `pdf-lib` codebase. Of course, I also understand if it is more than you're interested in doing.

Here are some other useful bits of `pdfkit` code:
* https://github.com/foliojs/pdfkit/blob/master/lib/font/standard.coffee#L19-L32
* https://github.com/foliojs/pdfkit/blob/master/lib/font/afm.coffee#L87-L94
* https://github.com/foliojs/pdfkit/blob/master/lib/mixins/text.coffee#L322-L329

Here are some of the relevant sections of the [PDF specification](https://github.com/Hopding/pdf-lib/blob/master/pdf_specification.pdf):
* **7.3.4.2:** Literal Strings
* **7.3.4.3:** Hexadecimal Strings
* **9.6.2:** Type 1 Fonts
* **Annex D:** Character Sets and Encodings

jerp:
I committed a version that uses the mapping of PDFKit. I did not check this mapping but being a user of this library for many years, I trust it better than myself.

The way i implemented it, requires to send a PDFHexString instance to drawText instead of a string. This Hex String is encoded properly. See example where I modified the following lines:
 
     const [helveticaFontRef, helveticaFont] = pdfDoc.embedStandardFont(HELVETIVA_FONT);
     ...
     helveticaFont.encode('Olé! - Œ')[0]

`embedStandardFont` now returns a tuple where the second argument is a text encoder, in this case, `helveticaFont`.

`helveticaFont.encode` should return a tuple, the first element is the encoded text. I believe, a second argument should give information about position advances (see pdfkit). 

Although, I'm an experience Javascript developer, I know little about Typescript, so your input is most than welcome.

The next thing I'm looking into is Font subsetting. This is an important requirement for me (file size).

Hopding:
Wow, this is great Jerome! You churned this out very quickly. I'm impressed.

I'll do a full review later this evening. It looks good overall. There are a few stylistic and linting related things that will need tweaked. I'd also like for the `drawText` operator to automatically encode text if possible. But I haven't fully thought through how that might work. Perhaps there are reasons this would be difficult that I haven't considered?

P.S. You may have already found this, but the easiest way to try out code changes is by running the integration tests with [`yarn it:node`](https://github.com/Hopding/pdf-lib/blob/master/package.json#L17). You can control which of the tests get run by commenting out the exports of tests you don't wish to run in the [index.ts](https://github.com/Hopding/pdf-lib/blob/master/__integration_tests__/tests/index.ts) file.

jerp:
I didn't know `yarn it:node`, I like it! All tests passed.

About the above, implementing it within `drawText`, would require this function to know what font is being used. For now, it only knows about the font name given at page level. When subsetting fonts, this gets more important since glyphs need to be collected prior to encode the font.

Hopding:
What do you think about passing the font object to the `drawText` operator, and encoding text "behind the scenes"?

For example, this is how it works right now:
```javascript
const [helveticaFontRef, helveticaFont] = pdfDoc.embedStandardFont(HELVETIVA_FONT); 
// ...
drawText(helveticaFont.encode('Olé! - Œ')[0], {
  x: PAGE_1_WIDTH * 0.5 -  30,
  y: PAGE_1_HEIGHT - 48 - 30,
  font: HELVETIVA_FONT,
  size: 12,
}),
```
But it could potentially work like this:
```javascript
const [helveticaFontRef, helveticaFont] = pdfDoc.embedStandardFont(HELVETIVA_FONT); 
// ...
drawText('Olé! - Œ', {
  x: PAGE_1_WIDTH * 0.5 -  30,
  y: PAGE_1_HEIGHT - 48 - 30,
  font: HELVETIVA_FONT,
  fontObject: helveticaFont,
  size: 12,
}),
```

My primary concern here is not technical. Clearly, it works either way. I'm just thinking about the API design. It's a balancing act between creating APIs that are too low-level and confusing for people to quickly grok and use, vs creating a super slick high-level API that requires more effort to maintain. 

In this case, neither of these APIs is super elegant and easy to use. However, I think passing the font object will be easier to document, and users will be less likely to make mistakes this way. It also paves the way for supporting automatic text wrapping by the `drawText` operator, because it would now have access to the font object required to measure strings.

I'd like to hear your thoughts on this. And of course, I'm open to alternative solutions as well.

Hopding:
Can you please run `yarn lint` on this as well? This will check the files to ensure they comply with the project's linting standards. It will also automatically fix any formatting changes that it can, so be sure to commit any changes you've made before running it.

P.S. You can prevent the linter from reformatting a block of code by adding a `// prettier-ignore` comment. But this should be used very sparingly. The linter is usually right.

src/core/pdf-structures/factories/PDFStandardFontFactory.ts
```diff
@@ -0,0 +1,118 @@
...
+  constructor(fontName: IStandard14FontsUnion) {
+    validate(
+      fontName,
+      oneOf(...Standard14Fonts),
+      'PDFDocument.embedStandardFont: "fontName" must be one of the Standard 14 Fonts: ' +
+        values(Standard14Fonts).join(', '),
+    );
+    this.fontName = fontName
+  }
+  encodeText(text: string): PDFHexString {
```
Hopding: For the most part, classes in `pdf-lib` use arrow function methods, e.g.
```js
class Foo {
  // Normal method
  bar() { ... }

  // Arrow Function Method
  bar = () => {}
}
```

There are a couple of benefits to this, one of which is that arrow function methods support implicit returns. So the above method can become:
```js
encodeText = (text: string): PDFHexString =>
  PDFHexString.fromString(text
    .split('')
    .map((char: string): number => char.charCodeAt(0))
    .map(toWinAnsi)
    .map((charCode: number): string => charCode.toString(16))
    .join(''))
```

Either approach works correctly. But using arrow functions with implicit returns (where possible) is more in-line with other `pdf-lib` classes.

It is also unnecessary to type the arguments and returns types of arrow functions. TypeScript will infer these types for you:
```js
encodeText = (text: string): PDFHexString =>
  PDFHexString.fromString(text
    .split('')
    .map(char => char.charCodeAt(0))
    .map(toWinAnsi)
    .map(charCode => charCode.toString(16))
    .join(''))
```

src/core/pdf-structures/factories/PDFStandardFontFactory.ts
```diff
@@ -0,0 +1,118 @@
...
+		return PDFHexString.fromString(text
+		.split('')
+		.map((char: string): number => char.charCodeAt(0))
+		.map(toWinAnsi)
+		.map((charCode: number): string => charCode.toString(16))
+		.join(''))
+	}
+  encode(text: string): [PDFHexString] {
```
Hopding: Can this be removed for now, and the `encodeText` method be renamed to just `encode`? I read your comment that:
> a second argument should give information about position advances (see pdfkit).

However, since this isn't currently implemented (indeed, the font metrics files to obtain this info don't currently exist in the project) I'd prefer to leave this method out for now.

jerp: I was going to suggest the opposite, keep the signature like it is. But I'm ok with your suggestion too.
Will work on this on Monday (and the rest).

src/core/pdf-structures/factories/PDFStandardFontFactory.ts
```diff
@@ -0,0 +1,118 @@
...
+import { isInstance, validate, oneOf } from 'utils/validate';
+import values from 'lodash/values';
+
+import Standard14Fonts, {
+  IStandard14FontsUnion,
+} from 'core/pdf-document/Standard14Fonts';
+
+import PDFFontEncoder from 'core/pdf-structures/factories/PDFFontEncoder'
+
+const toWinAnsi= (charCode: number): number => {
```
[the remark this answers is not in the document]
jerp: I was a bit puzzled by index signature message, but now, I did it like this:

```typescript
    const UnicodeToWinAnsiMap:{ [index:number] : number } = {
        402: 131, // ƒ
        ...,
    }
    const toWinAnsi= (charCode: number): number => UnicodeToWinAnsiMap[charCode]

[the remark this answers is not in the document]
Comment (Hopding) [reply src/core/pdf-structures/factories/PDFStandardFontFactory.ts]: I think you'll also need to include the default value (`|| charCode`) if the map doesn't contain a mapping?
```js
const toWinAnsi = (charCode: number): number => UnicodeToWinAnsiMap[charCode] || charCode
```

Also, while your type definition for the `UnicodeToWinAnsiMap` is correct, there's no need to include it for a simple mapping object. TypeScript should infer the type for you:
```js
const UnicodeToWinAnsiMap = {
    402: 131, // ƒ
    ...,
}
```
Type inference is one of TypeScript's strong points and it helps avoid a lot of type definition boilerplate. It puts languages like Java or C++ to shame 😄

src/core/pdf-structures/factories/PDFStandardFontFactory.ts
Hopding: I'd suggest including the type for the method arrow function's argument (I think it's required, actually), but not for the `.map` lambda arrow functions:
```js
// Include the param type here
encodeText = (text: string): PDFHexString =>
  PDFHexString.fromString(text
    .split('')
    // but not here
    .map(char => char.charCodeAt(0))
    .map(toWinAnsi)
    // or here
    .map(charCode => charCode.toString(16))
    .join(''))
```

jerp: Got it! Thanks

jerp:
Sure for the default charCode, unit tests detected it! 

About inference, got error messages, but I’ll give it another go.

src/core/pdf-structures/factories/PDFStandardFontFactory.ts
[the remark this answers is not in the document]
jerp: I did try several options, I get an error message if I omit type definition of `UnicodeToWinAnsiMap`

see a simplified case on [TS Playground](https://www.typescriptlang.org/play/index.html#src=const%20a%20%3D%20(i%3A%20number)%3A%20number%20%3D%3E%20(%7B%201%3A%202%2C%203%3A%204%2C%7D)%5Bi%5D%0D%0A%0D%0Aconst%20B%3A%20%7B%20%5Bindex%3A%20number%5D%3A%20number%20%7D%20%3D%20%7B%201%3A%202%2C%203%3A%204%2C%20%7D%0D%0Aconst%20b%20%3D%20(i%3A%20number)%3A%20number%20%3D%3E%20B%5Bi%5D) 

When the flag `noImplicitAny` is on, this triggers an error message. This saw that this option is on in `tsconfig.json`

Hopding:
Just tried defining that `UnicodeToWinAnsiMap` without a type definition myself and got an error as well. I wouldn't worry about trying to get TS to infer it. Doesn't seem like that will work in this case.

Hopding:
@jerp Is there anything else you plan to do on this PR before it's merged? If not, then I'll go ahead and merge it.

jerp:
If you’re statisfied with the latest change go ahead. I’m looking at
fontkit, font subsetting and how to lighten it all up. Fancy a font-lib
repo?

Hopding:
Yeah, this looks great! I'll go ahead and merge it. I still think I might want to change the API for the `drawText` operator, but that doesn't need to be done in this PR. I'll just look at doing it before I cut the next release.

Thanks again for the work you did on this! It's a real help to me, and provides a much needed feature for `pdf-lib`. Congrats on being the first contributor to the project other than myself!  🎉 

-----

Regarding `fontkit`: That dependency is a _pretty_ heavy one (see https://github.com/Hopding/pdf-lib/issues/14 and https://github.com/Hopding/pdf-lib/issues/13). I was working on decreasing its bundle size and publishing my forked version to npm a few weeks ago. Right now, `pdf-lib` points directly at my `fontkit` fork's GitHub repo. Also, [the build](https://github.com/Hopding/fontkit/blob/master/package.json#L18) is rather icky and brittle, which contributes to the size problem. I've been planning to pick that back up soon, but other issues have kept me from it. The branch I was working on is [here](https://github.com/Hopding/fontkit/tree/CleanupBuilds).

The reason I forked it (and the [`unicode-properties`](https://github.com/Hopding/unicode-properties) repo) was to make the library run in the browser without using browserify (see https://github.com/foliojs/unicode-properties/pull/2). I would have preferred to just make a PR to the original repo and avoid forking, but the maintainer didn't seem to be open to the changes I made. Though perhaps this is changing (see https://github.com/foliojs/unicode-properties/pull/5)?

Are you planning on making any changes in particular? Or still just exploring the codebase and what it supports?

Issue #37, opened 2018-10-12
Error when parsing some PDF documents


gregbacchus:
Version: pdf-lib@0.3.0

Unfortunately I'm unable to provide the offending documents as they come from a customer.

```
TypeError: Cannot read property '0' of null
  at parseObjData (.../node_modules/pdf-lib/lib/core/pdf-parser/parseObjectStream.js:44:64)
  at Object.parseObjectStream [as default] (.../node_modules/pdf-lib/lib/core/pdf-parser/parseObjectStream.js:67:19)
  at Object.exports.default (.../node_modules/pdf-lib/lib/core/pdf-parser/parseStream.js:82:55)
  at Object.parseDictOrStream [as default] (.../node_modules/pdf-lib/lib/core/pdf-parser/parseDictOrStream.js:28:44)
  at Object.parseIndirectObj [as default] (.../node_modules/pdf-lib/lib/core/pdf-parser/parseIndirectObj.js:43:41)
  at parseBodySection (.../node_modules/pdf-lib/lib/core/pdf-parser/parseDocument.js:24:48)
  at Object.parseDocument [as default] (.../node_modules/pdf-lib/lib/core/pdf-parser/parseDocument.js:78:21)
  at PDFParser.parse (.../node_modules/pdf-lib/lib/core/pdf-parser/PDFParser.js:38:40)
  at Function.PDFDocumentFactory.load (.../node_modules/pdf-lib/lib/core/pdf-document/PDFDocumentFactory.js:45:35)
  at FileModel.<anonymous> (.../src/models/file.model.ts:128:40)
```

Hopding:
Hello @gregbacchus. Has `pdf-lib` worked for you on other documents? I’m wondering if it has failed consistently with every document you’ve tried, or if it’s just this particular one that’s generating an error.

gregbacchus:
Yes, it has worked on other documents from other sources. I have only just started playing with `pdf-lib` so haven't tested it on very many examples, but I find that the documents that we receive are from many different sources (and I suspect that have varying degrees of compliance with PDF specs).

Hopding:
This will be difficult to debug since I can't directly access the documents, and have been unable to reproduce this myself. If you're interested in helping, can you please do the following:

1. Open `node_modules/pdf-lib/lib/core/pdf-parser/parseObjectStream.js` in an editor
2. Add 3 log statements to the `parseObjData` function, defined on line `28`:
```js
var parseObjData = function (dict, input) {
[...]

PR #38, opened 2018-10-17
fix for issue #37, new lines between datums

gregbacchus:
This is to fix handling of datums being separated by new lines.


Hopding:
Thanks for working on this @gregbacchus!

I checked the spec to see exactly what formats are allowed for the integer pairs at the beginning of an object stream. 

Section 7.5.7 of the [PDF specification](https://github.com/Hopding/pdf-lib/blob/master/pdf_specification.pdf) contains the following:

> **7.5.7 Object Streams**
> ...
> The stream data in an object stream shall contain the following items:
> * **N** pairs of integers separated by white space, where the first integer in each pair shall represent the object number of a compressed object and the second integer shall represent the byte offset in the decoded stream of that object, relative to the first object stored in the object stream, the value of the stream's first entry. The offsets shall be in increasing order.
>
> ...

Note that it states the integer pairs are to be **separated by white space**. The current implementation of `parseObjData` only works if the integer pairs are separated by space (` `) characters. However, in PDF Land, the term "whitespace" actually refers to 6 different characters.

Section 7.2.2 of the [PDF specification](https://github.com/Hopding/pdf-lib/blob/master/pdf_specification.pdf) explains whitespace further:

> **7.2.2 Character Set**
>
> The PDF character set is divided into three classes, called regular, delimiter, and white-space characters. This classification determines the grouping of characters into tokens. The rules defined in this sub-clause apply to all characters in the file except within strings, streams, and comments.
>
> The White-space characters shown in Table 1 separate syntactic constructs such as names and numbers from each other. All white-space characters are equivalent, except in comments, strings, and streams. In all other contexts, PDF treats any sequence of consecutive white-space characters as one character.
>
> **Table 1 – White-space characters**
>
> | Decimal | Hexadecimal | Octal | Name                 | JS Regex Literal |
> |---------|-------------|-------|----------------------|------------------|
> | 0       | 00          | 000   | Null (NUL)           | `\0`             |
> | 9       | 09          | 011   | HORIZONTAL TAB (HT)  | `\t`             |
> | 10      | 0A          | 012   | LINE FEED (LF)       | `\n`             |
> | 12      | 0C          | 014   | FORM FEED (FF)       | `\f`             |
> | 13      | 0D          | 015   | CARRIAGE RETURN (CR) | `\r`             |
> | 32      | 20          | 040   | SPACE (SP)           | ` `              |
(Note that I added the "JS Regex Literal" column)

-----

All of this is to say that your current fix gets the `parseObjData` function closer to parsing all valid integer pair formats, but not quite all the way there. 

For example, the following integer pairs format is valid, but the `objDatumRegex` fails to match it:
```
1 


   2   3 4
```
Changing the `objDatumRegex` definition to the following seems to do the trick:
```
const objDatumRegex = /^[\0\t\n\f\r ]*(\d+)[\0\t\n\f\r ]*(\d+)[\0\t\n\f\r ]*/
```

Would you mind updating your PR with this? 

Also, if you can update the new object stream you added (`__tests__/core/pdf-parser/data/object-stream3`) to contain a few of these more complex integer pair formats, I'd greatly appreciate it! (You'll need to change the values of the `First` and `Length` dict entries if you add more characters to separate the integer pairs).

src/core/pdf-parser/parseBool.ts
```diff
@@ -18,21 +18,23 @@ const parseBool = (
...
 ): [PDFBoolean, Uint8Array] | void => {
-  const boolRegex = /^(?:[\n|\r| ]*)(true|false)((?= |\]|\n|\r))?/;
+  const trimmed = trimArray(input);
+  const boolRegex = /^(?:[\0\t\n\f\r ]*)(true|false)((?=[\0\t\n\f\r \]]))?/;
 
   // Search for first character that isn't part of a boolean
   let idx = 0;
-  while (String.fromCharCode(input[idx]).match(/^[ \n\rtruefalse]/)) idx += 1;
+  while (String.fromCharCode(trimmed[idx]).match(/^[\0\t\n\f\r truefalse]/) && idx < trimmed.length) idx += 1;
+  console.log('parseBool', idx, encodeURI(arrayToString(trimmed, 0, 100)));
```
Hopding:
```suggestion
```
I assume this was just for debugging purposes?

src/core/pdf-parser/parseNumber.ts
```diff
@@ -19,11 +19,12 @@ const parseNumber = (
   { onParseNumber }: IParseHandlers = {},
 ): [PDFNumber, Uint8Array] | void => {
   const trimmed = trimArray(input);
-  const numRegex = /^(((\+{1}|-{1})?\d+(\.\d+)?)|((\+{1}|-{1})?\.\d+))/;
+  const numRegex = /^(([+-]?\d+(\.\d+)?)|([+-]?\.\d+))/;
 
   // Search for the first character that isn't part of a number
   let idx = 0;
   while (String.fromCharCode(trimmed[idx]).match(/^[+-.\d]/)) idx += 1;
+  console.log('parseNumber', idx, arrayToString(trimmed, 0, 100));
```
Hopding:
```suggestion
```
I assume this was just for debugging purposes?

src/core/pdf-parser/parseBool.ts
gregbacchus: Yes, sorry, shouldn't have been committed.

src/core/pdf-parser/parseNumber.ts
gregbacchus: Yes, sorry, shouldn't have been committed.

Issue #20, opened 2018-08-10
Draw Text with fixed Width


ithillel-aminev:
Hi, guys. That would be great if you to add drawing text with fixed width so that the text would wrap the next lines.

Hopding:
Hello @ithillel-aminev. I agree that this would be useful. It's a feature I plan implement in `pdf-lib` eventually. 

In the mean time, however, you should be able to accomplish this manually if you're willing to do a bit of of work. I've explained how to center and right-align text on a page here: https://github.com/Hopding/pdf-lib/issues/10. In that thread, I include code for measuring a string of text. Once you're able to do this, it should be easy enough to manually wrap your lines when they exceed your desired line length.

Please let me know if this doesn't work for you, or you have additional questions!

ithillel-aminev:
Thank you a lot

Hopding:
Happy to help!

I'm going to close this for now. Feel free to reopen it if you have additional questions 👍

ithillel-aminev:
Now I'm trying to wrap a text into lines and experience a problem with setting \n elements (\r or \r\n) in my test string literal. The result is the string is not wrapped and the \n elements are just drawn instead.

Hopding:
Version [`0.6.0`](https://www.npmjs.com/package/pdf-lib/v/0.6.0) is now published. It adds new methods to measure the width and height of text in standard and embedded fonts ([example](https://github.com/Hopding/pdf-lib/blob/master/examples/measuring_text/node/index.js)). The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v0.6.0).

You can install this new version with npm:
```
npm install pdf-lib@0.6.0
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@0.6.0/dist/pdf-lib.js
* https://unpkg.com/pdf-lib@0.6.0/dist/pdf-lib.min.js

rehman-ahmad-nbs:
> Version [`0.6.0`](https://www.npmjs.com/package/pdf-lib/v/0.6.0) is now published. It adds new methods to measure the width and height of text in standard and embedded fonts ([example](https://github.com/Hopding/pdf-lib/blob/master/examples/measuring_text/node/index.js)). The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v0.6.0).

Example link isn't working

Hopding:
[...]

PR #41, opened 2018-10-23
Special chars in name

gregbacchus:
Added unit tests for examples of names from pdf spec document. Added decoding of `#20` style encoded chars in name.


src/core/pdf-parser/parseName.ts
```diff
@@ -35,8 +35,12 @@ const parseName = (
   if (!result) return undefined;
 
   const [fullMatch, name] = result;
+  const decoded = name.replace(
```
Hopding: I think this decoding is a good idea. I'm not sure if the parsing function is the best spot to do it though. What do you think about moving this logic into the constructor of the [`PDFName`](https://github.com/Hopding/pdf-lib/blob/master/src/core/pdf-objects/PDFName.ts#L15) object itself? This object already has to encode characters in its [`toString()`](https://github.com/Hopding/pdf-lib/blob/master/src/core/pdf-objects/PDFName.ts#L45-L55) method. So I think it makes sense to have the decoding logic in there too.

gregbacchus: I had originally wondered about that... I guess I wasn't sure if having the encoding in the `.toString()` is actually ideal. Maybe better would be to make a `PDFName.fromEncoded()` as there is already a `PDFName.from()` in use. Thoughts?

Hopding: Sure, creating a `PDFName.fromEncoded()` and placing the decoding logic there makes sense to me. That way it's possible to create a `PDFName` object without going through the decoding process, if needed.

PR #92, opened 2019-04-11
Add browser example

fatso83:
This is a first draft at making a browser example. I say this, because _I thought there might be a better way_ of sharing existing code and assets. As you can see from the code, it is almost identical to the node version, exept for the preamble and postamble, which is platform specific. I could extract the common bits into a separate module and have two init modules - one for node and one for the browser, put them all into one project and share all the assets (instead of bloating the repo by duplicating them - although I am unsure if Git's own compression avoids this in praxis).


fatso83:
Here's the combined version, with a common module: https://github.com/Hopding/pdf-lib/compare/master...fatso83:browser-and-node-example.

Just for discussion. Both have their merits. Having them split at least makes each simpler.

examples/document_creation/browser/package.json
```diff
@@ -0,0 +1,42 @@
+{
```
Hopding: I'd like for the examples to be as lightweight as possible, and using `browserify`, `watchify`, and `babel` seems pretty noisy. Many users of `pdf-lib` will not be familiar with these tools, and having them in the examples adds significant extra overhead for these users.

What do you think about just having an `index.html` and `index.js` file for this example? (No `npm` or `package.json`). We could use the [UMD builds from unpkg](https://github.com/Hopding/pdf-lib#umd-module) via a `<script>` tag. This should remove the need for `browserify`. And as for Babel, I don't think we need it at all. As far as I'm aware, all modern browsers support the ES6 syntax (async/await, destructuring, etc...) used in the `index.js` file.

fatso83: Sure, I'll take a stab at it. Seems like module support is equally on par with `async/await`, so I could put the user scripts in a single file.

Btw, should I go for the shared main module approach (the latter) or separate folders?

Hopding: I think I’d favor keeping them split, since that makes each example a bit more straightforward.

But I’m also open to trying to share just the assets. If you want to try that, I’m curious what it’d look like with a simplified browser version of the code.

Hopding:
@fatso83 Any updates on this? 

(There's no rush - just making sure you're still planning to work on this 😄)

fatso83:
Sorry - totally forgot about this. I don't have time until mid-June, though.

Hopding:
No worries

philipjmurphy:
I modified @fatso83's browser index.js file and just included it via require.js and it works very well on Chrome and Safari. Also, used the download link code instead as the original example appears to stream the content directly into the web page. I think this is what @Hopding was getting at. Thought I'd just send on as might help. To run just start a http server in folder e.g. http-server -o (to install npm -g i http-server)

[pdf-lib-require.zip](https://github.com/Hopding/pdf-lib/files/3283500/pdf-lib-require.zip)

Hopding:
v1.0.0 was just released. It includes multiple browser examples (inspired by the work in this PR), so I think it's reasonable to close this PR now.

PR #97, opened 2019-04-25
Filter may be a PDFArray with a single item

### LINKED ISSUE #96: Cannot decode FlatDecode Object Stream
Hi,

I have an issue while parsing some pdfs (example here : https://s3.amazonaws.com/mle-sandbox/_Q3+ssol.pdf)

Line 128 of parseStream.ts `dict.getMaybe('Filter')` returns a PDF Array with a single item.

I will make a pull request that fix my case (to be discussed ;) )

mlecoq:
To fix  #96


mlecoq:
@Hopding done !

Hopding:
Looks great. Thanks again @mlecoq!

PR #108, opened 2019-05-13
handles scientific notations

### LINKED ISSUE #107: Unable to load a pdf file after adding a circle
After drawing on one document, I cannot load it then.

In my document there are boxes :

    <</BBox [-0.00000000000000224167 0.000000000000000239557 1.38095 2.7619]

*-0.00000000000000224167* is seen as a number in file _parseNumber.ts_

The following line:
         
    const pdfNumber = PDFNumber.fromString(num);

Transform the string to 

    2.24167e-15

While adding circles.

Then when I want to load the produced document (to combine it with others), the following regex (in parseNumber.ts) :
    
    const numRegex = /^(([+-]?\d+(\.\d+)?)|([+-]?\.\d+))/;

Can not handle the full string _2.24167e-15_ and only extract _2.24167_. 

Parsing stops on _e-15_

I will make a pull request

mlecoq:
To fix #107

### DIFF
--- a/__tests__/core/pdf-objects/PDFNumber.spec.ts
+++ b/__tests__/core/pdf-objects/PDFNumber.spec.ts
@@ -1,12 +1,5 @@
-import PDFObjectIndex from 'core/pdf-document/PDFObjectIndex';
-import {
-  PDFIndirectObject,
-  PDFIndirectReference,
-  PDFNumber,
-  PDFNumber,
-  PDFString,
-} from 'core/pdf-objects';
-import { typedArrayFor } from 'utils';
+import { PDFNumber } from 'core/pdf-objects';
+import { toCharCode, typedArrayFor } from 'utils';
 
 describe(`PDFNumber`, () => {
   it(`requires a number to be constructed`, () => {
@@ -66,5 +59,47 @@ describe(`PDFNumber`, () => {
       pdfNumber.copyBytesInto(buffer);
       expect(buffer).toEqual(typedArrayFor('9000'));
     });
+  });
+
+  it(`can be converted to a string`, () => {
+    expect(String(PDFNumber.fromNumber(21))).toEqual('21');
+    expect(String(PDFNumber.fromNumber(-43))).toEqual('-43');
+    expect(String(PDFNumber.fromNumber(3.403e38))).toEqual(
+      '340300000000000000000000000000000000000',
+    );
+    expect(String(PDFNumber.fromNumber(-3.403e38))).toEqual(
+      '-340300000000000000000000000000000000000',
+    );
+    expect(String(PDFNumber.fromNumber(-3.403e-38))).toEqual(
+      '-0.000000000000000000000000000000000000034030000000000005',
+    );
+  });
+
+  it(`can provide its size in bytes`, () => {
+    expect(PDFNumber.fromNumber(21).bytesSize()).toBe(2);
+    expect(PDFNumber.fromNumber(-43).bytesSize()).toBe(3);
+    expect(PDFNumber.fromNumber(3.403e38).bytesSize()).toBe(39);
+    expect(PDFNumber.fromNumber(-3.403e38).bytesSize()).toBe(40);
+  });
+
+  it(`can be serialized`, () => {
+    const buffer1 = new Uint8Array(8).fill(toCharCode(' '));
+    PDFNumber.fromNumber(21).copyBytesInto(buffer1);
+
+    expect(buffer1).toEqual(typedArrayFor('21      '));
+
+    const buffer2 = new Uint8Array(40).fill(toCharCode(' '));
+    PDFNumber.fromNumber(-3.403e38).copyBytesInto(buffer2);
+    expect(buffer2).toEqual(
+      typedArrayFor('-340300000000000000000000000000000000000'),
+    );
+
+    const buffer3 = new Uint8Array(64).fill(toCharCode(' '));
+    PDFNumber.fromNumber(-3.403e-38).copyBytesInto(buffer3);
+    expect(buffer3).toEqual(
+      typedArrayFor(
+        '-0.000000000000000000000000000000000000034030000000000005       ',
+      ),
+    );
   });
 });
--- a/src/core/pdf-objects/PDFNumber.ts
+++ b/src/core/pdf-objects/PDFNumber.ts
@@ -20,20 +20,18 @@ class PDFNumber extends PDFObject {
   };
 
   number: number;
-  private stringValue: string;
 
   constructor(num: number) {
     super();
     validate(num, isFinite, 'Can only construct PDFNumbers from Numbers');
     this.number = num;
-    this.stringValue = numberToString(num);
   }
 
   clone = () => PDFNumber.fromNumber(this.number);
 
-  toString = (): string => this.stringValue;
+  toString = (): string => numberToString(this.number);
 
-  bytesSize = () => this.stringValue.length;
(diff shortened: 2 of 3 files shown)


Hopding:
Great work @mlecoq! 

I am currently working on a rewrite of `pdf-lib` to improve performance and clean up some APIs. While working on the rewrite of `PDFNumber`, I ran into this scientific notation scenario. I double checked the spec, and numbers in PDF files should *not* contain exponential notation.

But unfortunately, JavaScript doesn’t allow you to disable scientific notation when converting a number to a string. So it requires some manual conversions. Here’s the relevant code and tests I wrote a few days ago:

* https://github.com/Hopding/pdf-lib/blob/Rewrite/src/core/objects/PDFNumber.ts#L14
* https://github.com/Hopding/pdf-lib/blob/Rewrite/src/utils/numbers.ts#L6
* https://github.com/Hopding/pdf-lib/blob/Rewrite/tests/core/objects/PDFNumber.spec.ts#L18

I think this would be the best solution, rather than updating the parser to handle exponential notation.

That being said, have you found that PDF readers are able to tolerate numbers with exponential notation? If so, perhaps we need to do both (not write out exponential notation, but also update the parser to tolerate it as input).

mlecoq:
I have juste used _Preview_ on my Mac to check the result and It worked with the current code inside 

/BBox [ -2.24167e-15 2.39557e-16 1.38095 2.7619 ]

mlecoq:
With Chrome and Firefox it is fined also but there is an issue on adobe reader

mlecoq:
I will update my pull request with your suggestion, I need to be sure that most of pdf readers supports generated files. Thank you for your quick feedback !

mlecoq:
After adding the code you provided, when merging several document, only the first one is displayed

mlecoq:
The issue comes from numberToString method which does not always return exact value. For example:

    numberToString(Number('-1.23356e-15'))

returns 

    '-0.0000000000000012335600000000001'

instead of 

    '-0.00000000000000123356'


Using this lib: https://github.com/shrpne/from-exponential

solves my issue 


I will update my PR

mlecoq:
I thought it solves the issue but in fact not 😕

Hopding:
So, you're trying to merge several PDFs (none of which contain exponential notation), and the resulting PDF (which doesn't contain exponential notation) can't be opened?

mlecoq:
It can be opened. When I look at the file size, pdfs seem to be concatenated but only first one is displayed.

mlecoq:
Here is the file with first fix (scientific notation embedded): all pages are visible in your browser (firefox in my case)

[https://s3.amazonaws.com/mle-sandbox/file-scientific.pdf](scientific-notation)

With _numberToString_ or _from-exponential_ only the first page is visible

[https://s3.amazonaws.com/mle-sandbox/file.pdf](without-scientific)

The beginning of the files are similar (except Bbox statement) but at the end, two files are different. I cannot explain why for the moment.

mlecoq:
Here are the 3 files I am concatenating:

[](https://s3.amazonaws.com/mle-sandbox/file-1.pdf)
[](https://s3.amazonaws.com/mle-sandbox/file-2.pdf)
[](https://s3.amazonaws.com/mle-sandbox/file-3.pdf)

To concatenate 2 files, I just load them, then use _addPage_ method on one of them to add pages of the other

mlecoq:
If you have any idea ! It's hard to understand how a small modification in PDFNumber has these side-effects.

Hopding:
Just using `numberToString` seems to work fine for me. 

I applied these specific changes to the `master` branch:
```diff
diff --git a/src/core/pdf-objects/PDFNumber.ts b/src/core/pdf-objects/PDFNumber.ts
index b3a4bf9..e7d482d 100644
--- a/src/core/pdf-objects/PDFNumber.ts
+++ b/src/core/pdf-objects/PDFNumber.ts
@@ -6,6 +6,30 @@ import { validate } from 'utils/validate';
 
 import PDFObject from './PDFObject';
 
+export const numberToString = (num: number) => {
+  let numStr = String(num);
+
+  if (Math.abs(num) < 1.0) {
+    const e = parseInt(num.toString().split('e-')[1]);
+    if (e) {
+      const negative = num < 0;
+      if (negative) num *= -1;
+      num *= Math.pow(10, e - 1);
+      numStr = '0.' + new Array(e).join('0') + num.toString().substring(2);
+      if (negative) numStr = '-' + numStr;
+    }
+  } else {
+    let e = parseInt(num.toString().split('+')[1]);
+    if (e > 20) {
+      e -= 20;
+      num /= Math.pow(10, e);
+      numStr = num.toString() + new Array(e + 1).join('0');
+    }
+  }
+
+  return numStr;
+};
+
 class PDFNumber extends PDFObject {
   static fromNumber = (num: number) => new PDFNumber(num);
 
@@ -28,9 +52,9 @@ class PDFNumber extends PDFObject {
 
   clone = () => PDFNumber.fromNumber(this.number);
 
-  toString = (): string => this.number.toString();
+  toString = (): string => numberToString(this.number);
 
-  bytesSize = () => this.toString().length;
+  bytesSize = () => numberToString(this.number).length;
 
   copyBytesInto = (buffer: Uint8Array): Uint8Array =>
     addStringToBuffer(this.toString(), buffer);
```

After that, I ran this simple script:
```js
const fs = require('fs');

const pdfDoc1 = PDFDocumentFactory.load(fs.readFileSync('file1.pdf'));
const pdfDoc2 = PDFDocumentFactory.load(fs.readFileSync('file2.pdf'));
const pdfDoc3 = PDFDocumentFactory.load(fs.readFileSync('file3.pdf'));

const pdfDoc = PDFDocumentFactory.create();
pdfDoc.addPage(pdfDoc1.getPages()[0]);
pdfDoc.addPage(pdfDoc2.getPages()[0]);
pdfDoc.addPage(pdfDoc3.getPages()[0]);

const pdfbytes = PDFDocumentWriter.saveToBytes(pdfDoc, { useObjectStreams: false });
```

The result is this merged PDF, which opens fine in Mac Preview, Acrobat, and Firefox: [merged.pdf](https://github.com/Hopding/pdf-lib/files/3174963/tmp.pdf)

Please try out the changes I made in the diff and let me know what you find!

mlecoq:
I will try soon, thanks !

mlecoq:
@Hopding after adding these modifications: https://github.com/Hopding/pdf-lib/commit/3c5cd8f2cda4855b1aa7e9225076a80abda35a19 

it works fine ! 

I don't understand why using _stringValue_ attribute leads to an error, have you any explanation ?

Thank you very much for your help ! (again 😉 )

Hopding:
@mlecoq I suspect the reason is that `this.number` is a public member of `PDFNumber`. This means that its value can be safely modified. E.g. this is safe to do in the current (`master`) version of `pdf-lib`:
```js
const pdfNum = PDFNumber.from(21);
pdfNum.number = 42;
```
When you added a `stringValue` property, you inadvertently made the above operation unsafe, because it would cause `this.number` and `this.stringValue` to get out of sync.

In the new version of `pdf-lib` that I'm working on, `this.number` and `this.stringValue` are private members. As such, it is not safe to modify either of them directly. That's why you saw `this.stringValue` as a property of `PDFNumber` in the code I referenced. So in the new version, you'd have to do:
```js
let pdfNum = PDFNumber.of(21);
pdfNum = PDFNumber.of(42);
```
(Note that the reason for this change is to speed up document serialization by only doing the `number -> string` conversion once for a given `PDFNumber` instance).

mlecoq:
Thanks for your feed-back !

PR #162, opened 2019-08-17
Encode PDFString containing non-Latin characters

PlushBeaver:
ASCII representation for PDFString cannot handle code points above 127 (7-bit ASCII). Such strings have to be encoded as Unicode (UTF16+BOM). This implementation opts to represent them as hex strings for ease of encoding.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pdf-lib",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Create and modify PDF files with JavaScript",
   "author": "Andrew Dillon <andrew.dillon.j@gmail.com>",
   "contributors": [
@@ -56,7 +56,8 @@
     "@pdf-lib/standard-fonts": "^0.0.4",
     "pako": "^1.0.10",
     "png-ts": "^0.0.3",
-    "tslib": "^1.10.0"
+    "tslib": "^1.10.0",
+    "utfx": "^1.0.1"
   },
   "devDependencies": {
     "@pdf-lib/fontkit": "^0.0.4",
--- a/src/core/objects/PDFString.ts
+++ b/src/core/objects/PDFString.ts
@@ -1,37 +1,51 @@
 import PDFObject from 'src/core/objects/PDFObject';
-import CharCodes from 'src/core/syntax/CharCodes';
-import { copyStringIntoBuffer } from 'src/utils';
+import { copyStringIntoBuffer, toHexStringOfMinLength } from 'src/utils';
+import utfx = require('utfx');
 
 class PDFString extends PDFObject {
   // The PDF spec allows newlines and parens to appear directly within a literal
   // string. These character _may_ be escaped. But they do not _have_ to be. So
   // for simplicity, we will not bother escaping them.
-  static of = (value: string) => new PDFString(value);
+  static of = (value: string) => new PDFString(value, false);
 
   private readonly value: string;
 
-  private constructor(value: string) {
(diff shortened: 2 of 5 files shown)


Hopding:
Hello @PlushBeaver! I just had a chance to review this. My apologies for the late response.

Thank you very much for the time and effort you spent on this! I have a couple of questions/concerns:

1. The `PDFString` class is meant to represent PDF literal strings, specifically. The [`PDFHexString`](https://github.com/Hopding/pdf-lib/blob/master/src/core/objects/PDFHexString.ts) class exists to represent hex strings. So I would prefer not to have a `PDFString` class encode itself as a hex string.
However, the PDF spec does allow for literal strings to contain characters outside the ASCII set via octal character codes. E.g. `(Foo\053Bar)` represents `Foo+Bar`. What do you think about implementing things this way?

2. PDF files should not contain non-ascii characters within string literals (this would violate the spec). And checking to see if they do whenever a `PDFString` is created will negatively impact parsing performance (which is already a very intensive operation). So if we adopt these changes, we'll need a way to ensure they do not run when creating strings during document parsing.

3. Aside from checking for non-ascii characters when parsing, what use cases are you thinking of that these changes would address?

PlushBeaver:
Thanks for the concerns raised, @Hopding.

1. I find the need for library users to choose between `PDFString` and `PDFHexString` impractical, because it exposes PDF internals and limitations (while providing flexibility if needed). Would you recommend using `PDFHexString` when there's any chance that the string contains characters outside ASCII? If so, this PR should not exist at all.

    Otherwise, I must indeed take care of octal representation when decoding strings and consider using it when encoding. Hex seems preferable for non-English text (Cyrillic, CJK, etc), while octal may be better for seldom special characters in mostly Latin text (diacritics, punctuation, etc).

2. Parsing got completely off my mind (see below). Are parsing tests used as benchmarks? Note that decoding needs not be done while parsing, it can be a lazy operation on a string stored "as is" (encoded).

3. Primary use case is inserting Unicode metadata and outline entries. I [stumbled](https://gitlab.pagedmedia.org/tools/pagedjs-cli/issues/8) into this with pagedjs-cli. That tool uses an outdated version of pdf-lib and it uses `PDFString`s. However, you raised an interesting question of loading documents with such data.

Hopding:
Hello @PlushBeaver! My apologies for the delayed response - I've been quite busy. But I haven't forgotten about this. I've been thinking about the best way to handle all of this. I finally made some decisions, and have been working on implementing things in #204 over the past few days. The changes I made in #204 are all based on your work here, but with a few differences. Please take a look at #204 and let me know if I missed anything that you solved here. If it looks good to you, then I'll merge it (closing both #162 and #204) and the changes will go out in the next pdf-lib release!

PlushBeaver:
Hello, @Hopding. Specialized facilities from #204 are both handy and flexible enough not only to solve metatata encoding issues, but also to implement Unicode outline using new `PDFHexString.fromText(string)`. This PR can indeed be closed in favor of #204. Thank you for the excellent library you provide!

Hopding:
Version [`1.2.0`](https://www.npmjs.com/package/pdf-lib/v/1.2.0) is now published. It contains the changes from this #204. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.2.0).

```
npm install pdf-lib@1.2.0
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@1.2.0/dist/pdf-lib.js
* https://unpkg.com/pdf-lib@1.2.0/dist/pdf-lib.min.js

PR #173, opened 2019-08-27
Draw wrapped lines of text

multiplegeorges:
A simple wrapper function around `drawText` that automatically wraps lines at a specified length.

### DIFF
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -20,6 +20,7 @@ import {
   PDFPageDrawRectangleOptions,
   PDFPageDrawSquareOptions,
   PDFPageDrawTextOptions,
+  PDFPageDrawWrappedTextOptions,
 } from 'src/api/PDFPageOptions';
 import { degrees, Rotation, toDegrees } from 'src/api/rotations';
 import { StandardFonts } from 'src/api/StandardFonts';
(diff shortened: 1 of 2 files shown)


scottie-schneider:
This is exactly what I need...

multiplegeorges:
@scottie-schneider It's very easy to monkeypatch into your project from this PR. 

You may want to do that since I suspect this might take a while to merge because I haven't included any testing. I was waiting to get some feedback before going ahead with that.

scottie-schneider:
@multiplegeorges Got it working, thanks to your examples and code. Thanks sir!

Hopding:
Hello @multiplegeorges! Thank you very much for working on this. It's a feature that's been on my mind but that I haven't had the time to work on. I think it will be a great addition that many users will find very helpful.

Here are my thoughts after looking the PR over:
* I'm not sure about the API. Specifically, I don't think I like the addition of the `drawWrappedText` method. I **do** like the `maxWidth` option though. What do you think about removing the `drawWrappedText` method and adding the `maxWidth` to the existing `drawText` method? If the user doesn't specify `maxWidth`, we can just default to the existing behavior. But if they do specify it, then we can branch into the code you've written to do the line wrapping.
* I'm concerned about the efficiency of the implementation. There are 2 opportunities I can see for improving this:
  1. Reuse `candidateString` instead of re-running `words.slice(startIndex, i).join(' ')`.
  2. Measuring the width of a string can be a somewhat expensive operation. Perhaps we could measure the width of each word and sum their widths to compute the full line's width. This would allow the word widths to be cached (within or across `drawText` calls).

  However, I would be interested in having some concrete numbers on how long it takes to compute the layout before and after the optimizations. I think we want optimization `#1` for sure. But optimization `#2` is a bit more involved, and I wouldn't want to add that complexity if it doesn't provide tangible performance benefits. Would you be willing to collect these numbers for large blocks of text with standard and embedded fonts?

vicary:
Splitting by whitespaces may prove difficult for languages that don't use space as word separators (i.e. Chinese), and in rare cases, a very long Latin word.

I suggest a word breaking character `wordBreak = "-"` as the default for English users, while Chinese users can put an empty string for their specific use cases.

I can do the word-breaking logic in a separate PR on top of this.

Site-node: @Hopding Word-splitting, especially in Chinese paragraphs, may render pre-calculated word widths in suggestion #2 a waste of resources though. What do you think?

Hopding:
Thanks again for everybody's work on this. I've merged a variant of this code in https://github.com/Hopding/pdf-lib/pull/198. It will go out in the next release of pdf-lib. If anybody would like to document the `maxWidth` and `wordBreak` options in the README or API docs, that'd be awesome. These options are currently undocumented.

Hopding:
Version [`1.1.1`](https://www.npmjs.com/package/pdf-lib/v/1.1.1) is now published. It contains the `maxWidth` and `wordBreak` options for the `PDFPage.drawText()` method. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.1.1).

```
npm install pdf-lib@1.1.1
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@1.1.1/dist/pdf-lib.js
* https://unpkg.com/pdf-lib@1.1.1/dist/pdf-lib.min.js

Issue #186, opened 2019-09-09
Issue while copying a specific pdf


mlecoq:
Hi,

I have an issue with the following pdf :
https://mle-sandbox.s3.amazonaws.com/1526.pdf

When I append it to an existing pdf, it fails due to an undefined value. 

The code I use to append it 

```

    const firstPlan = await PDFDocument.load(fs.readFileSync(`${__dirname}/pdf/firstPlan.pdf`), {
        parseSpeed: ParseSpeeds.Fastest
    });

  
        const doc = await PDFDocument.load(`${__dirname}/pdf/unsupportedPlan.pdf`, {
            parseSpeed: ParseSpeeds.Fastest
        });

        const pages = await firstPlan.copyPages(doc, [...doc.getPages().keys()]);

        for (let page of pages) {
            firstPlan.addPage(page);
        }

        fs.writeFileSync(`${__dirname}/results/result.pdf`, await firstPlan.save());

```
 
Making some modifications in the lib fix the issue. I will make a pull request soon

Hopding:
Version [`1.1.1`](https://www.npmjs.com/package/pdf-lib/v/1.1.1) is now published. It contains the fix for this issue. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.1.1).

You can install this new version with npm:
```
npm install pdf-lib@1.1.1
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@1.1.1/dist/pdf-lib.js
* https://unpkg.com/pdf-lib@1.1.1/dist/pdf-lib.min.js

PR #187, opened 2019-09-09
support undefined value while copying

mlecoq:
To fix https://github.com/Hopding/pdf-lib/issues/186

The idea is to support undefined object which has been encountered on a specific pdf file


Hopding:
Hello @mlecoq! Great work this. I implemented a slightly different version of your fix in #199 and added a regression test.

Issue #114, opened 2019-05-23
Missing support for embedding SVG


lepidotteri:
It could be a very handy feature imo, right now, in a browser environment for example, you would need to parse the SVG in order to be drawn into a canvas only for the newly drawn image base64 to be taken.

Would it be possible to add this kind of feature by re-using the existing vector graphics drawing functions of pdf-lib?

Hopding:
Hello @lepidotteri! This feature could definitely be added to `pdf-lib`. It would be implemented, as you suggest, by parsing the SVG and invoking `pdf-lib`'s vector graphics functions for each of the SVG's elements. [`pdfkit` does exactly this](https://github.com/foliojs/pdfkit/blob/master/lib/path.js#L409-L414) to implement it's SVG feature.

I can't commit to delivering this feature by any specific date. There are several other important things I'm planning to work on first. But this feature is certainly on my radar. And, of course, I'm always open to somebody submitting a PR for this!

jlmessenger:
@lepidotteri - FYI, there is now SVG path support in version `1.2.0`.
https://pdf-lib.js.org/docs/api/classes/pdfpage#drawsvgpath

So if you had an SVG like:
```xml
<svg width="4cm" height="4cm" viewBox="0 0 400 400"
     xmlns="http://www.w3.org/2000/svg" version="1.1">
  <title>Example triangle01- simple example of a 'path'</title>
  <desc>A path that draws a triangle</desc>
  <rect x="1" y="1" width="398" height="398"
        fill="none" stroke="blue" />
  <path d="M 100 100 L 300 100 L 200 300 z"
        fill="red" stroke="blue" stroke-width="3" />
</svg>
```

You could render that in pdf-lib with this code:
```js
const pdfDoc = await PDFDocument.create();
const page = pdfDoc.addPage(PageSizes.Letter);

// 1" from upper left
const positionX = 72;
const positionY = 720;

const svgWidth = 398;
const svgHeight = 398;
const blue = rgb(0, 0, 1);
const red = rgb(1, 0, 0);

page.drawRectangle({
	x: positionX,
	y: positionY - svgHeight,
	width: svgWidth,
	height: svgHeight,
	borderColor: blue,
});
page.drawSvgPath('M 100 100 L 300 100 L 200 300 z', {
	x: positionX,
	y: positionY,
	color: red,
	borderColor: blue,
	borderWidth: 3
});
```
[...]

PR #202, opened 2019-09-26
Add page.drawSVG(path) method

jlmessenger:
Brought over the SVG path methods from `pdfkit`, with attribution and updated to TypeScript.

Addresses issue: https://github.com/Hopding/pdf-lib/issues/114

Here's a sample test page with the output with variations on the options.
https://gist.github.com/jlmessenger/c8323fcb7891f30843e59a391cae925a

### DIFF
--- a/apps/node/index.ts
+++ b/apps/node/index.ts
@@ -6,6 +6,7 @@ import readline from 'readline';
 import test1 from './tests/test1';
 import test10 from './tests/test10';
 import test11 from './tests/test11';
+import test12 from './tests/test12';
 import test2 from './tests/test2';
 import test3 from './tests/test3';
 import test4 from './tests/test4';
@@ -120,7 +121,7 @@ const main = async () => {
     // prettier-ignore
     const allTests = [
     test1, test2, test3, test4, test5, test6, test7, test8, test9, test10,
-    test11
+    test11, test12
   ];
 
     const tests = testIdx ? [allTests[testIdx - 1]] : allTests;
--- a/apps/node/tests/test12.ts
+++ b/apps/node/tests/test12.ts
@@ -1,0 +1,100 @@
+import { Assets } from '..';
+import {
+	PageSizes,
+	PDFDocument,
+	rgb
+} from '../../..';
+
+const inchToPt = (inches: number) => Math.round(inches * 72);
+
+export default async (_assets: Assets) => {
+	const pdfDoc = await PDFDocument.create();
+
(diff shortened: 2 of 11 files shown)


jlmessenger:
@Hopding let me know what you think.

Hopding:
Hello @jlmessenger! This is certainly a feature that I'd like to get into pdf-lib. So thank you for working on this. I'm planning to do a thorough review within the next week. However, I've looked over this briefly and a few things pop out at me:
* This new code will need to have some unit tests added before it can be merged. Primarily `src/api/svgPath.ts`.
* I also try to ensure that all new API changes are tested in the three tester apps (in the `apps/` dir) to prevent regressions.
* I do not think these changes fully address #114, because these changes support drawing SVG *paths*. They do not allow SVG *files* to be embedded. It certainly does not have to be done as part of this PR, but I'm curious how much effort you think it would take to support SVG embedding?
* Along the lines of my previous note, I think it would be better to call the new method `PDFPage.drawSvgPath(...)`.

As I mentioned, I'll try to do a more thorough review of the code itself soon. Please let me know if you have any questions/concerns!

Hopding:
I think the best way to test the `src/api/svgPath.ts` module would be to have a series of tests cases that call `svgPathOperators` with various path strings and then verify that the resulting array contains the correct operators and arguments in the correct sequence. The key thing would be to exercise all the different paths/functions through the code in `src/api/svgPath.ts`. I would like to have nearly 100% coverage for code like this (similar to how the PDF parser code has almost 100% test coverage). You can generate a coverage report with [`yarn testc`](https://github.com/Hopding/pdf-lib/blob/master/package.json#L21).

I hope this helps. Please let me know if you have any additional questions!

jlmessenger:
@Hopding - Thanks for the direction.
I used the [SVG path specification](https://svgwg.org/svg2-draft/paths.html) as a source for path samples, and added unit tests.

Hopding:
Version [`1.2.0`](https://www.npmjs.com/package/pdf-lib/v/1.2.0) is now published. It contains the `PDFPage.drawSvgPath(...)` method. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.2.0).

```
npm install pdf-lib@1.2.0
```
It's also available on unpkg:
* https://unpkg.com/pdf-lib@1.2.0/dist/pdf-lib.js
* https://unpkg.com/pdf-lib@1.2.0/dist/pdf-lib.min.js

Issue #122, opened 2019-06-11
Get fields by page


sdfereday:
Hi all,
just wondering if it's possible to return a list of fields for each individual page? At present I've no problems getting a hold of the different pages and fields separately but curious if they have any sort of page/field relationship so they can be grouped together somehow.

For example, TextBox1 belongs to Page1, TextBox2 belongs to Page2, etc.

If that makes sense...

Cheers!

jboxxx:
Hi - I am also curious about linking pages with fields, otherwise it feels like I'm flying blind.  How are you at present listing out all of the fields on your pdf?   Thanks

sdfereday:
> Hi - I am also curious about linking pages with fields, otherwise it feels like I'm flying blind. How are you at present listing out all of the fields on your pdf? And for that matter, have you figured out how to change the values on them? Trying to find in the docs rn but haven't had luck so far. Thanks

@jboxxx Have a sneaky peak at this for filling in form fields:
[https://github.com/Hopding/pdf-lib/issues/48](url)

As for the pages at present, I've basically saved out an individual PDF for each page, then when I've filled in each one, I merge them together at the end as one whole PDF. It's not ideal but seems to work for now. :)

Hopding:
Hello @sdfereday! 

There is indeed such a relationship. A given form field belongs to a particular page if the form field's ref is in that page's `Annots` array.

I created an example script to demonstrate how this works:
```js
// ...imports omitted...

const getAcroFieldRefs = (pdfDoc) => {
  if (!pdfDoc.catalog.getMaybe('AcroForm')) return [];
  const acroForm = pdfDoc.index.lookup(pdfDoc.catalog.get('AcroForm'));

  if (!acroForm.getMaybe('Fields')) return [];
  const acroFields = pdfDoc.index.lookup(acroForm.get('Fields'));

  return acroFields.array;
};

const pdfDoc = PDFDocumentFactory.load(fs.readFileSync('./form.pdf'));

const fieldRefs = getAcroFieldRefs(pdfDoc);

console.log('REFS:', fieldRefs.map(String));

const pages = pdfDoc.getPages();

const page1AnnotsRaw = pdfDoc.index.lookupMaybe(pages[0].getMaybe('Annots'));
const page1Annots = page1AnnotsRaw ? page1AnnotsRaw.array : [];

const page2AnnotsRaw = pdfDoc.index.lookupMaybe(pages[1].getMaybe('Annots'));
[...]

Issue #111, opened 2019-05-14
How to fill form radio fields


pammann-work:
Can pdf-lib be used to programmatically fill out radio fields in a template PDF?
Analog to #48

PDF source looks like:
```
...
148 0 obj
<</BC[]/BG[0.952896 0.952896 0.952896]>>
endobj
149 0 obj
<</Count 0/Type/Outlines>>
endobj
150 0 obj
[214 0 R 171 0 R 184 0 R 202 0 R 176 0 R 186 0 R 180 0 R 192 0 R 188 0 R 178 0 R 196 0 R 198 0 R 200 0 R 194 0 R 190 0 R 182 0 R 151 0 R 152 0 R 153 0 R 18 0 R 70 0 R 19 0 R 24 0 R 79 0 R 71 0 R 17 0 R 75 0 R 68 0 R 67 0 R 59 0 R 74 0 R 37 0 R 86 0 R 38 0 R 90 0 R 94 0 R 88 0 R 40 0 R 89 0 R 30 0 R 28 0 R 34 0 R 26 0 R 105 0 R 113 0 R 99 0 R 127 0 R 104 0 R 142 0 R 98 0 R 143 0 R 115 0 R 110 0 R 147 0 R 117 0 R 141 0 R 130 0 R 103 0 R]
endobj
151 0 obj
<</FT/Btn/Ff 49152/Kids[211 0 R]/Opt 156 0 R/T(Radio 1)>>
endobj
152 0 obj
<</FT/Btn/Ff 49152/Kids[208 0 R]/Opt 155 0 R/T(Radio 2)>>
endobj
153 0 obj
<</FT/Btn/Ff 49152/Kids[204 0 R]/Opt 154 0 R/T(Radio 3)>>
endobj
154 0 obj
[(Choice1)]
endobj
155 0 obj
[(Choice1)]
endobj
156 0 obj
[(Choice1)]
endobj
157 0 obj
...
```

Hopding:
Hello @pammann-work! 

I created an example script demonstrating how to do this: [pdf-lib_radio_button_form_fill_example.zip](https://github.com/Hopding/pdf-lib/files/3195143/pdf-lib_radio_button_form_fill_example.zip)

You can run the script with the following commands (you may have to tweak them a bit if you're running Windows):
```
unzip pdf-lib_radio_button_form_fill_example.zip
cd pdf-lib_radio_button_form_fill_example
npm install
node index.js
```

[...]

Issue #109, opened 2019-05-14
How to fill form checkbox fields


pammann-work:
Can pdf-lib be used to programmatically fill out checkbox fields in a template PDF?
Analog to #48

Hopding:
Hello @pammann-work! 

I created an example script demonstrating how to do this: [pdf-lib_check_box_form_fill_example.zip](https://github.com/Hopding/pdf-lib/files/3195157/pdf-lib_check_box_form_fill_example.zip)

You can run the script with the following commands (you may have to tweak them a bit if you're running Windows):
```
unzip pdf-lib_check_box_form_fill_example.zip
cd pdf-lib_check_box_form_fill_example
npm install
node index.js
```

Here's the input PDF: [simple_form.pdf](https://github.com/Hopding/pdf-lib/files/3195159/simple_form.pdf)
And here's the output: [filled_form.pdf](https://github.com/Hopding/pdf-lib/files/3195160/filled_form.pdf)

Here's a snippet from the script showing the gist of the process:
```js
const getAcroFields = pdfDoc => {
  if (!pdfDoc.catalog.getMaybe('AcroForm')) return [];
  const acroForm = pdfDoc.index.lookup(pdfDoc.catalog.get('AcroForm'));

  if (!acroForm.getMaybe('Fields')) return [];
  const acroFields = pdfDoc.index.lookup(acroForm.get('Fields'));

  return acroFields.array.map(pdfDoc.index.lookup);
};

const findAcroFieldByName = (pdfDoc, name) => {
  const acroFields = getAcroFields(pdfDoc);
  return acroFields.find(acroField => {
    const fieldName = acroField.getMaybe('T');
    return !!fieldName && fieldName.string === name;
  });
};

const pdfDoc = PDFDocumentFactory.load(fs.readFileSync('./simple_form.pdf'));

const fooCheckBox = findAcroFieldByName(pdfDoc, 'FooCheckBox');
const barCheckBox = findAcroFieldByName(pdfDoc, 'BarCheckBox');
const quxCheckBox = findAcroFieldByName(pdfDoc, 'QuxCheckBox');
const bazCheckBox = findAcroFieldByName(pdfDoc, 'BazCheckBox');

console.log('Selecting "Foo" and "Qux" boxes...');
fooCheckBox.set('V', PDFName.from('Yes'));
quxCheckBox.set('V', PDFName.from('Yes'));

console.log('Saving modified PDF...');
[...]

Issue #48, opened 2018-11-22
How to fill form text fields


Hopding:
Can `pdf-lib` be used to programmatically fill out text fields in a template PDF?

Hopding:
Yes, this is possible. No high level API currently exists for it, so the code is a bit ugly. But it works all the same!

Here's an example Node script that fills out a [DoD character sheet](http://dnd.wizards.com/articles/features/character_sheets):
```js
const getAcroFields = (pdfDoc) => {
  if (!pdfDoc.catalog.getMaybe('AcroForm')) return [];
  const acroForm = pdfDoc.index.lookup(pdfDoc.catalog.get('AcroForm'));

  if (!acroForm.getMaybe('Fields')) return [];
  const acroFields = pdfDoc.index.lookup(acroForm.get('Fields'));

  return acroFields.array.map(pdfDoc.index.lookup);
};

const findAcroFieldByName = (pdfDoc, name) => {
  const acroFields = getAcroFields(pdfDoc);
  return acroFields.find((acroField) => {
    const fieldName = acroField.getMaybe('T');
    return !!fieldName && fieldName.string === name;
  });
};

const fillAcroTextField = (
  pdfDoc,
  acroField,
  fontObject,
  text,
  fontSize = 15,
) => {
  const fieldRect = acroField.get('Rect');
  const fieldWidth = fieldRect.get(2).number - fieldRect.get(0).number;
  const fieldHeight = fieldRect.get(3).number - fieldRect.get(1).number;

  const appearanceStream = pdfDoc.register(
    PDFContentStream.of(
      PDFDictionary.from({
        Type: PDFName.from('XObject'),
        Subtype: PDFName.from('Form'),
        BBox: PDFArray.fromArray([
          PDFNumber.fromNumber(0),
          PDFNumber.fromNumber(0),
          PDFNumber.fromNumber(fieldWidth),
          PDFNumber.fromNumber(fieldHeight),
        ], pdfDoc.index),
        Resources: PDFDictionary.from({
          Font: PDFDictionary.from({
            FontObject: fontObject
          }, pdfDoc.index)
        }, pdfDoc.index),
      }, pdfDoc.index),
      drawLinesOfText(text.split('\n'), {
        x: 2,
        y: fieldHeight - 13,
        font: 'FontObject',
        size: fontSize,
        colorRgb: [0, 0, 0],
      })
    ),
  );

  acroField.set('V', PDFString.fromString(text));
[...]

PR #254, opened 2019-12-15
Test stream

cshenks:
Adding some tests for the base Stream class

### DIFF
--- a/tests/core/streams/Stream.spec.ts
+++ b/tests/core/streams/Stream.spec.ts
@@ -44,34 +44,29 @@ describe(`Stream`, () => {
 
     describe('can get the next n bytes', () => {
         let stream:Stream;
+        const bytes = Uint8Array.from([1, 2, 258]);
 
         beforeEach(() => {
-            stream = new Stream(Uint8Array.from([1, 2, 258]));
+            stream = new Stream(bytes);
         });
 
         describe('when zero length is specified', () => {
             it('and the values are not clamped', () => {
-                const bytes = stream.getBytes(0, false);
-                expect(bytes.length).toEqual(stream.length);
-                expect(bytes[0]).toEqual(1);
-                expect(bytes[1]).toEqual(2);
-                expect(bytes[2]).toEqual(2);
+                expect(stream.getBytes(0, false)).toEqual(bytes);
             });
 
             it('and the values are clamped', () => {
-                const bytes = stream.getBytes(0, true);
-                expect(bytes.length).toEqual(stream.length);
-                expect(bytes[0]).toEqual(1);
-                expect(bytes[1]).toEqual(2);
-                expect(bytes[2]).toEqual(2);
+                const gottenBytes = stream.getBytes(0, true);
+                expect(gottenBytes instanceof Uint8ClampedArray).toEqual(true);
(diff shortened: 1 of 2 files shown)


tests/core/streams/Stream.spec.ts
```diff
@@ -0,0 +1,128 @@
+import Stream from 'src/core/streams/Stream';
```
Hopding: Would you mind running `yarn lint`? That will update this file to comply with the project's formatting standards. You can then just commit and push the changes.

tests/core/streams/Stream.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(stream.getUint16()).toEqual(514);
+    });
+
+    it('can get thirty-two bit integers', () => {
+        const bytes = Uint8Array.from([1, 1, 1, 1, 2, 2, 2, 2, 255, 255, 255, 255]);
+        const stream = new Stream(bytes);
+        expect(stream.getInt32()).toEqual(16843009);
+        expect(stream.getInt32()).toEqual(33686018);
+        expect(stream.getInt32()).toEqual(-1);
+    });
```
Hopding: This test confused me a bit at first. I initially thought the `-1` was returned because the end of the stream had been reached. But it's actually just because that's what `[255, 255, 255, 255]` evaluates to as a signed 32 bit integer. 

Can we swap around the order of the byte subsequences to eliminate the potential for confusion here? For example:
```js
it('can get thirty-two bit integers', () => {
  const bytes = Uint8Array.from([1, 1, 1, 1, 255, 255, 255, 255, 2, 2, 2, 2]);
  const stream = new Stream(bytes);
  expect(stream.getInt32()).toEqual(16843009);
  expect(stream.getInt32()).toEqual(-1);
  expect(stream.getInt32()).toEqual(33686018);
});
```

tests/core/streams/Stream.spec.ts
cshenks: Updated

tests/core/streams/Stream.spec.ts
cshenks: Updated

cshenks:
@Hopding Back to you. Thanks for being so prompt!

On an unrelated note, I'm working on an Acroforms API. Is there any work begun on this that isn't yet in a branch?

Hopding:
@cshenks I'm glad to hear you're interested in working on this! Creating an Acroforms API has been near the top of my unofficial pdf-lib roadmap for some time now. I just haven't had the time to get around to it myself yet. I'd be happy to collaborate on this with you. I have some high-level ideas about how to best implement such an API in pdf-lib.

While there is not currently any work being done on an Acroforms API, you might mind the following issues to be helpful. They demonstrate how to create/edit Acroforms with the low-level APIs:
* https://github.com/Hopding/pdf-lib/issues/39
* https://github.com/Hopding/pdf-lib/issues/48
* https://github.com/Hopding/pdf-lib/issues/109
* https://github.com/Hopding/pdf-lib/issues/111
* https://github.com/Hopding/pdf-lib/issues/122

cshenks:
@Hopding I'd be happy to collaborate! Is there any chance you'd be willing to give me permissions to add a new project for this?

Issue #39, opened 2018-10-20
How to get/set Annot or Widget from pdf


ssleptsov:
Hey, is it possible to get/set Annot or Widget for existing pdf file?
I'm trying to replace this pdfkit code with your library 

```js
const signature = pdf.ref({
  Type: 'Sig',
  Filter: 'Adobe.PPKLite',
  SubFilter: 'adbe.pkcs7.detached',
  ByteRange: [
    0,
    DEFAULT_BYTE_RANGE_PLACEHOLDER,
    DEFAULT_BYTE_RANGE_PLACEHOLDER,
    DEFAULT_BYTE_RANGE_PLACEHOLDER,
  ],
  Contents: Buffer.from(String.fromCharCode(0).repeat(signatureLength)),
  Reason: new String(reason), // eslint-disable-line no-new-wrappers
  M: new Date(),
});

// Generate signature annotation widget
const widget = pdf.ref({
  Type: 'Annot',
  Subtype: 'Widget',
  FT: 'Sig',
  Rect: [0, 0, 0, 0],
  V: signature,
  T: new String('Signature1'), // eslint-disable-line no-new-wrappers
  F: 4,
  P: pdf._root.data.Pages.data.Kids[0], // eslint-disable-line no-underscore-dangle
});

// Include the widget in a page
pdf._root.data.Pages.data.Kids[0].data.Annots = [widget];

// Create a form (with the widget) and link in the _root
const form = pdf.ref({
  SigFlags: 3,
  Fields: [widget],
});
pdf._root.data.AcroForm = form;
```
Thanks!

Hopding:
Hello @ssleptsov. This is certainly possible, as `pdf-lib` provides low level access to pdf objects. However, there is not currently a nice high level API for manipulating annotations or acroform objects, so you'll have to put up with some verbose boilerplate for now :smile:

Here's how to do it:
```js
const pdfBytes = /* A Uint8Array containing the PDF you wish to modify */
const pdfDoc = PDFDocumentFactory.load(pdfBytes);

const [FontHelvetica] = pdfDoc.embedStandardFont('Helvetica');

// Define the signature dictionary (Detailed in Table 252 of PDF specification)
// Note that the dict below isn't actually valid. I just made up its `ByteRange`
// and `Contents` entries. It's just an example.
const signatureDict = PDFDictionary.from({
  Type: PDFName.from('Sig'),
  Filter: PDFName.from('Adobe.PPKLite'),
[...]

PR #255, opened 2019-12-17
Acroforms base

cshenks:
To begin adding general AcroForms functionality and address some issues listed below, I've added a `PDFAcroForm` structure. Right now it only returns an extension of `PDFDict` with getters for the attributes listed on pg. 431 of the PDF spec. It also dereferences the members of the AcroForm fields array before returning them. 

I plan to add a `PDFAcroFormField` structure next to make this more usable.
@Hopding could we add a project board for AcroForms work? I'd be happy to manage it.

### Relevant issues
- #39
- #48
- #109
- #111
- #122


Hopding:
@cshenks I created an a project board for this work: https://github.com/Hopding/pdf-lib/projects/1. Unfortunately, GitHub doesn't have a very flexible permissions model for personal (non-organization) repos. So I can't grant you access to just manage the board. However, I added Kanban automation, so statuses should update automatically. I think the only management that will be necessary is to add cards to the board. Let me know if this is what you were looking for!

src/core/objects/PDFDict.ts
```diff
@@ -45,6 +46,8 @@ class PDFDict extends PDFObject {
     return this.dict.has(key);
   }
 
+  lookupMaybe(key: PDFName): PDFObject | undefined;
```
Hopding: I think this method is redundant, given that looking up a key with no type specified already returns an optional value: 

https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/objects/PDFDict.ts#L66

src/core/objects/PDFDict.ts
```diff
@@ -59,7 +62,7 @@ class PDFDict extends PDFObject {
   lookupMaybe(key: PDFName, type: typeof PDFRef): PDFRef | undefined;
   lookupMaybe(key: PDFName, type: typeof PDFString): PDFString | undefined;
 
-  lookupMaybe(key: PDFName, type: any) {
+  lookupMaybe(key: PDFName, type?: any) {
```
Hopding: See above.

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,59 @@
...
+  PDFName,
+  PDFNumber,
+  PDFObject,
+  PDFStream,
+  PDFString,
+} from 'src/core';
+
+class PDFAcroForm extends PDFDict {
+  static fromMapWithContext(
+    dict: Map<PDFName, PDFObject>,
```
Hopding: I think it would be best to use the `DictMap` type here, instead of redefining `Map<PDFName, PDFObject>`:

https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/objects/PDFDict.ts#L14

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,59 @@
...
+    context: PDFContext,
+  ): PDFAcroForm {
+    dict.set(
```
Hopding: This is a stylistic nit-pick, but I think an explicit existence check would be a little more clear. Also, explicit construction of PDF primitive objects is discouraged in favor of the `context.obj(...)` helper, which allows you to construct PDF objects with literal syntax:

https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/PDFContext.ts#L141-L146

For example:
```js
if (!dict.has(PDFName.of('Fields'))) {
  dict.set(PDFName.of('Fields'), context.obj([]))
}
```

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  get Fields(): PDFArray {
+    const fieldRefs = this.get(PDFName.of('Fields')) as PDFArray;
+    return fieldRefs.map((fieldRef) => this.context.lookup(fieldRef, PDFDict));
+  }
+
+  get NeedsAppearances(): PDFBool | undefined {
+    return this.get(PDFName.of('NeedsAppearances')) as PDFBool;
+  }
```
Hopding: I can certainly see the appeal of using getter methods here. However, for the sake of API consistency I would prefer that they be normal methods (see [here](https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/structures/PDFCatalog.ts#L21), [here](https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/structures/PDFPageTree.ts#L28), and [here](https://github.com/Hopding/pdf-lib/blob/620747a044c2b3864c9d844b7c67b924b4f94ea5/src/core/structures/PDFPageLeaf.ts#L71)).

Also, these seem like perfect opportunities to use typed lookups instead of static casts. For example:
```js
return this.lookupMaybe(PDFName.of('NeedsAppearances'), PDFBool);
```

cshenks: @Hopding  I suppose I don't fully understand the relationship between the contents of a `PDFDict` and its corresponding `PDFContext`. Is the presumption that every object that might be a value in a `PDFDict` has first been registered with the `PDFContext`?

cshenks:
> @cshenks I created an a project board for this work: https://github.com/Hopding/pdf-lib/projects/1. Unfortunately, GitHub doesn't have a very flexible permissions model for personal (non-organization) repos. So I can't grant you access to just manage the board. However, I added Kanban automation, so statuses should update automatically. I think the only management that will be necessary is to add cards to the board. Let me know if this is what you were looking for!

I'm having some trouble accepting your invitation to collaborate. I've seen other reports of invitations taking a while to work. Did you retract the invitation, by any chance?

src/core/structures/PDFAcroForm.ts
cshenks: Updated

src/core/structures/PDFAcroForm.ts
cshenks: > @Hopding I suppose I don't fully understand the relationship between the contents of a `PDFDict` and its corresponding `PDFContext`. Is the assumption that every value in a `PDFDict` has first been registered with the `PDFContext`? I'm not sure whether all objects need to be registered with a context or only indirect objects. If it's only indirect objects, the PDF spec indicates that these dictionary entries are directly accessible, so I think we could just use `get`.

Ah, nvm. I see how this works now. I'll make this change.

src/core/structures/PDFAcroForm.ts
cshenks: Updated

src/core/objects/PDFDict.ts
cshenks: Updated

src/core/objects/PDFDict.ts
cshenks: Updated

Hopding:
@cshenks I just resent the invite. Let me know if it works this time.

cshenks:
@Hopding The invite worked. Thanks!

Also if you'd like to cut a new `acroforms` branch and have this PR merged into that instead of `master`, let me know.

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,59 @@
...
+  PDFDict,
+  PDFName,
+  PDFNumber,
+  PDFStream,
+  PDFString,
+  XFAResource,
+} from 'src/core';
+import { DictMap } from 'src/core/objects/PDFDict';
+
+class PDFAcroForm extends PDFDict {
```
Hopding: This structure is going to be a bit problematic. Let me explain why:

Currently, the `PDFDict` subclass structures (`PDFCatalog`, `PDFPageLeaf`, and `PDFPageTree`) all have required `/Type` entries. This is important, because it allows us to detect what type of class a given `PDFDict` should be when we are parsing PDF files:

https://github.com/Hopding/pdf-lib/blob/8346da359c1c35089b2d02af07d29228e89482a3/src/core/parser/PDFObjectParser.ts#L194-L204

This means, for example, that the `PDFContext` for a `PDFDocument` will always contain `PDFPageTree` classes, not just `PDFDict` objects with `/Type /Pages`. So we could do this:

```js
const pdfDoc = PDFDocument.load(pdfBytes)
const pageTrees = pdfDoc
  .context
  .enumerateIndirectObjects()
  .find(([, obj]) => obj instanceof PDFPageTree);
```

But, unfortunately, the PDF specification does not define a required `/Type` for AcroForm dictionaries. This makes it very difficult to know whether or not a particular `PDFDict` should be converted to a `PDFAcroForm` when parsing a file. And if we can't create the `PDFAcroForm` instances when parsing, the only other place we can do that is when the `PDFCatalog.AcroForm` method is called. Of course, this would mean that the `PDFAcroForm` structure would simply be a wrapper around its underlying `PDFDict`. And `PDFAcroForm` objects could not be stored in the `PDFContext` object itself.

At best, this violates the convention set by `PDFCatalog`, `PDFPageLeaf`, and `PDFPageTree`. At worst, this makes the behavior of pdf-lib inconsistent and therefore difficult to reason about.

This issue is a bit difficult to explain. I hope I was able to communicate it clearly for you. Please let me know if you don't understand.

I'm not sure what the best way to handle this is. My initial reaction is that the approach pdf-lib currently uses to convert `PDFDict` objects to specialized types when parsing (checking their `/Type` entry) just isn't going to work going forward. And that it would be best to do away with that approach and adhere to a more flexible convention going forward. Of course, I also am cautious about making breaking API/behavior changes. So I'm still weighing the options. If you have any thoughts on the matter, I'd love to hear them!

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  protected constructor(map: DictMap, context: PDFContext) {
+    super(map, context);
+  }
+
+  Fields(): PDFArray {
+    const fieldRefs = this.lookup(PDFName.of('Fields'), PDFArray);
+    return fieldRefs.map((fieldRef) => this.context.lookup(fieldRef, PDFDict));
+  }
```
Hopding: All instances of `PDFObject` subclasses should exist as values in the `PDFContext` map or else be referenced by another `PDFObject` that is present in the `PDFContext`. Another way of saying this is that no `PDFObject` instances should be generated that are not going to be serialized when `PDFDocument.save()` is called. This is because `PDFObject` subclasses only exist to model objects in a PDF file. 

For example, it would be very easy for somebody to call `PDFAcroForm.Fields` and think they are getting the actual `Fields` array, since a `PDFArray` object is returned. And then they might modify that `PDFArray` thinking they are adding new fields that will be serialized when they save the document. But of course, this would not happen since the array they modified wasn't actually part of the `PDFDocument`.

The map being performed here violates this convention. The `PDFArray` returned by `PDFAcroForm.Fields` is an "orphan". It will not be serialized when `PDFDocument.save()` is invoked. This means that there's no reason for it to be a `PDFArray` in the first place. I think it would be best to just return a `PDFDict[]` instead.

Of course, if this method wasn't mapping and it just returned the value of the `/Fields` entry, it would make sense to return a `PDFArray`. Because it would be returning an object that actually exists in the PDF file.

An alternative to returning `PDFDict[]` here would be to create two methods, for example:
```js
Fields(): PDFArray {
  return this.lookup(PDFName.of('Fields'), PDFArray);
}

get fieldDicts(): PDFDict[] {
  const fieldRefs = this.Fields();
  const fieldDicts = new Array(fieldRefs.size());
  for (let idx = 0, len = fieldRefs.size(); idx < len; idx++) {
    fieldDicts[idx] = this.context.lookup(fieldRefs[idx], PDFDict);
  }
  return fieldDicts;
}
```

src/core/objects/PDFArray.ts
```diff
@@ -47,6 +47,14 @@ class PDFArray extends PDFObject {
     return this.array[index];
   }
 
+  map(mapFunc: (obj: PDFObject) => PDFObject): PDFArray {
+    const mappedPDFArray = new PDFArray(this.context);
+    this.array.forEach((obj) => {
+      mappedPDFArray.push(mapFunc(obj));
+    });
+    return mappedPDFArray;
+  }
```
Hopding: @cshenks 

I'm a big fan of declarative/functional programming and I'm a strong advocate for using `map`, `filter`,  and `forEach` over imperative `for` and `while` loops. However, when I was building pdf-lib I had to keep performance in mind (PDF processing is very intensive, especially for larger PDF files). So I decided to benchmark various ways of processing arrays in JavaScript: https://jsperf.com/array-testing-2. (see [jsperf_array_tests_chrome.pdf](https://github.com/Hopding/pdf-lib/files/3991917/jsperf_array_tests_chrome.pdf.pdf) and [jsperf_array_tests_firefox.pdf](https://github.com/Hopding/pdf-lib/files/3991918/jsperf_array_tests_firefox.pdf) for the results).

The results are rather unfortunate. `map`, `filter`, and `forEach` are several times slower than their `for` loop equivalents. So I decided to avoid using `map`, `filter`, and `forEach` in the pdf-lib codebase and to stick with raw `for` loops. It's a tough decision to make, because `for` loops are easy to mess up (off-by-one errors, anyone?) and are significantly less readable. However, I think the tradeoff is worth it since pdf-lib is sometimes used to process rather large PDF files.

So, given all of this, I would prefer to avoid introduced a `map` utility here and instead just use a `for` loop when needed. But if there is a common pattern that is used throughout the codebase (e.g. looking up each member of a `PDFArray`), we could introduce a helper on `PDFArray`. For example:
```js
lookupElements(): PDFObject[] {
  const values = new Array(this.size());
  for (let idx = 0, len = this.size(); idx < len; idx++) {
    values[idx] = this.context.lookup(this.get(idx));
  }
  return values;
}
```

src/core/structures/PDFAcroForm.ts
cshenks: Updated

src/core/objects/PDFArray.ts
cshenks: Updated

src/core/structures/PDFAcroForm.ts
cshenks: @Hopding I see the issue, but I'm afraid I probably don't know enough about the PDF spec to prescribe a great solution. Writing custom parsing logic for the Catalog seems the best way to go to me, and I think writing structure-level parsing logic would be moving toward the more flexible convention you mentioned. But please let me know if I'm misunderstanding.

Hopding: @cshenks Can you clarify what you mean by the following? I'm not quite sure I follow.
> I think structure-level (as opposed to object-level) parsing for the Catalog would be a step toward this, and would let us cover parsing the AcroForms dictionary separately from ordinary `PDFDict` structures. But please let me know if I'm misunderstanding.

Hopding: @cshenks 

After thinking about this for awhile, the simplest approach I can think of is to avoid extending `PDFDict` and to instead create wrappers for it. This way the parser doesn't need to know anything about the specialized wrapper types. It will just parse into `PDFDict`, `PDFArray`, etc... and leave it at that. Then, when a particular getter method is called, an instance of the specialized wrapper type can be instaniated (backed by the appropriate `PDFDict` instance). And, of course, these wrapper types would not be stored in the `PDFContext`.

I was curious how other PDF libraries handled this problem. (Of course, many PDF libraries only generate documents, so they can avoid this issue entirely). I took a look at the source code for [PDFBox](https://github.com/apache/pdfbox) and discovered that they use the same wrapper approach I described above. 

Here's their version of `PDFDict`: [`COSDictionary`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/cos/COSDictionary.java). Their parser [generates instances of this class](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdfparser/BaseParser.java#L197-L202
), without any sort of specialized type conversion/casting: 

```java
protected COSDictionary parseCOSDictionary() throws IOException
{
    readExpectedChar('<');
    readExpectedChar('<');
    skipSpaces();
    COSDictionary obj = new COSDictionary();
```

They have several `COSDictionary` wrapper classes in the [`pdmodel`](https://github.com/apache/pdfbox/tree/trunk/pdfbox/src/main/java/org/apache/pdfbox/pdmodel) directory. For example:

* [`PDDocumentCatalog`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocumentCatalog.java)
* [`PDPage`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDPage.java)
* [`PDAcroForm`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/PDAcroForm.java)

And they instantiate these wrappers when the corresponding getter methods are invoked:

**[`PDDocumentCatalog`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocument.java#L762-L778)**

```java
public PDDocumentCatalog getDocumentCatalog()
{
    if (documentCatalog == null)
    {
        COSDictionary trailer = document.getTrailer();
        COSBase dictionary = trailer.getDictionaryObject(COSName.ROOT);
        if (dictionary instanceof COSDictionary)
        {
            documentCatalog = new PDDocumentCatalog(this, (COSDictionary) dictionary);
        }
        else
        {
            documentCatalog = new PDDocumentCatalog(this);
        }
    }
    return documentCatalog;
}
```

**[`PDPage`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDPageTree.java#L228-L236)**

```java
public PDPage get(int index)
{
    COSDictionary dict = get(index + 1, root, 0);

    sanitizeType(dict);

    ResourceCache resourceCache = document != null ? document.getResourceCache() : null;
    return new PDPage(dict, resourceCache);
}
```

**[`PDAcroForm`](https://github.com/apache/pdfbox/blob/f9b02631102e841668a464c5699b7568cc8087f3/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocumentCatalog.java#L108-L116)**

```java
public PDAcroForm getAcroForm()
{
    if (cachedAcroForm == null)
    {
        COSDictionary dict = (COSDictionary)root.getDictionaryObject(COSName.ACRO_FORM);
        cachedAcroForm = dict == null ? null : new PDAcroForm(document, dict);
    }
    return cachedAcroForm;
}
```

So considering that we each arrived at a similar solution for the same problem independently, it seems likely to be the best solution, all things considered. Or at least better than the obvious alternatives.

The only complicating factor here for pdf-lib is that there is existing code that relies on the subclassing approach. And while converting this code to the wrapper approach would keep things consistent, it would also introduce a breaking change. So I think it would be best to leave the existing subclassing code as it is (for `PDFCatalog`, `PDFPageLeaf`, and `PDFPageTree`) and just use the wrapper approach for all new models.

Hopding:
@cshenks 
> Also if you'd like to cut a new `acroforms` branch and have this PR merged into that instead of `master`, let me know.

Yes, I think this would be preferable. I prefer to always keep `master` in a releasable state. And I would rather not release any `acroforms` changes until its built out further, since we might make breaking changes as the API evolves.

src/core/structures/PDFAcroForm.ts
cshenks: > @cshenks Can you clarify what you mean by the following? I'm not quite sure I follow.
> 
> > I think structure-level (as opposed to object-level) parsing for the Catalog would be a step toward this, and would let us cover parsing the AcroForms dictionary separately from ordinary `PDFDict` structures. But please let me know if I'm misunderstanding.

I think I was trying to say that if the document were parsed recursively, beginning with the root, it would be possible to parse the AcroForm dictionary based on its relation to the root without having a `/Type` field. This would not work with the existing parsing paradigm, but you could do another pass over the context after parsing to replace `root -> Catalog -> AcroForm` with a `PDFAcroForm` object, if one exists.

cshenks:
> @cshenks
> 
> > Also if you'd like to cut a new `acroforms` branch and have this PR merged into that instead of `master`, let me know.
> 
> Yes, I think this would be preferable. I prefer to always keep `master` in a releasable state. And I would rather not release any `acroforms` changes until its built out further, since we might make breaking changes as the API evolves.

Sounds good. If you could make this branch, I'll reassign the target branch for the merge

src/core/structures/PDFAcroForm.ts
cshenks: @Hopding This seems good to me. I'll update `PDFAcroForm` to wrap `PDFDict`

Hopding:
@cshenks Here's the branch: [`AcroFormsAPI`](https://github.com/Hopding/pdf-lib/tree/AcroFormsAPI).

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,61 @@
...
+  PDFString,
+} from 'src/core';
+
+class PDFAcroForm {
+  static fromDict(dict: PDFDict, context: PDFContext): PDFAcroForm {
+    if (!dict.has(PDFName.of('Fields'))) {
+      dict.set(PDFName.of('Fields'), context.obj([]));
+    }
+    return new PDFAcroForm(dict, context);
+  }
```
Hopding: It's not necessary to pass in a `PDFContext` here. Instead, we can simply use `dict.context`. I think this is preferable, actually, as it simplifies the API a bit and removes the risk of passing mismatched contexts.

For example:
```js
static fromDict(dict: PDFDict): PDFAcroForm {
  if (!dict.has(PDFName.of('Fields'))) {
    dict.set(PDFName.of('Fields'), dict.context.obj([]));
  }
  return new PDFAcroForm(dict);
}
```

src/core/objects/PDFArray.ts
```diff
@@ -47,6 +47,17 @@ class PDFArray extends PDFObject {
     return this.array[index];
   }
 
+  lookupElements(): Array<PDFObject | undefined>;
+  lookupElements(type: typeof PDFDict): PDFDict[];
+
+  lookupElements(type?: any) {
+    const values = new Array(this.size());
+    for (let idx = 0, len = this.size(); idx < len; idx++) {
+      values[idx] = this.context.lookup(this.get(idx), type);
```
Hopding: This is pedantic, but I think you can simplify this to the following:
```suggestion
      values[idx] = this.lookup(idx, type);
```

src/core/structures/PDFAcroForm.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  readonly context: PDFContext;
```
Hopding: See my previous comment about not needing to pass in a separate `PDFContext`.

Issue #241, opened 2019-11-20
Error: Unit Testing failed


punisher97:
Hi, i want contributing a little in this amazing project, but unit testing no passed 100%

- **OS:** Win10 x64
- **Node.js:** v10.16.3
- **Yarn:** 1.19.1
- **Git:** git version 2.22.0.windows.1

> Test Suites: 5 failed, 42 passed, 47 total
> Tests:       13 failed, 412 passed, 425 total
> Snapshots:   0 total
> Time:        108.816s
> Ran all test suites.

Output:

      15 |       const stream = new FlateStream(new Stream(encoded));
      16 | 
    > 17 |       expect(stream.decode()).toEqual(decoded);
         |                               ^
      18 |     });
      19 |   });
      20 | });

      at Object.<anonymous> (tests/core/streams/FlateStream.spec.ts:17:31)

  ● FlateStream › can decode flate encoded data (4)

    expect(received).toEqual(expected) // deep equality

    ]

      15 |       const stream = new FlateStream(new Stream(encoded));
      16 | 
    > 17 |       expect(stream.decode()).toEqual(decoded);
         |                               ^
      18 |     });
      19 |   });
      20 | });

      at Object.<anonymous> (tests/core/streams/FlateStream.spec.ts:17:31)

 FAIL  tests/core/streams/AsciiHexStream.spec.ts
  ● AsciiHexStream › can decode ascii hex encoded data (2)

    expect(received).toEqual(expected) // deep equality

On the other hand, for now adding QR code support just like pdfmake, with pdf-lib they have a very high quality! I would like to add linear gradient, what do you think?
thanks

cshenks:
@punisher97 If you're developing on Windows, it's likely that the stream text fixture line endings are being converted to Windows line endings during cloning. You might give this a shot.
`git config --global core.autocrlf false`.

punisher97:
@cshenks  Thanks i will try

Hopding:
@punisher97 Please let me know if @cshenks's suggestion fixes the issue for you. If it does, I'd like to document this in the [CONTRIBUTING.md](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md).

Hopding:
@punisher97 I'm glad you're interested in contributing to pdf-lib! However, I'm afraid I do not fully understand what you mean by the following:
[...]

PR #256, opened 2019-12-17
Add a note about Windows line endings for contributors

cshenks:
@Hopding The other option would be to add a `.gitattributes` that explicitly forbids line ending conversions during cloning. Let me know if you'd prefer that.

See https://github.com/Hopding/pdf-lib/issues/241

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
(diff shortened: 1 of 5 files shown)


Hopding:
If a `.gitattributes` file will solve the problem automatically, I'd prefer to do that rather than add notes to the `CONTRIBUTING.md` file. It would make for fewer steps and less opportunities for a contributor to accidentally mess up.

cshenks:
> If a `.gitattributes` file will solve the problem automatically, I'd prefer to do that rather than add notes to the `CONTRIBUTING.md` file. It would make for fewer steps and less opportunities for a contributor to accidentally mess up.

@Hopding updated and confirmed that the tests pass with a fresh clone.

Issue #197, opened 2019-09-24
Creating big PDF - JavaScript heap out of memory


MikeMnD:
It's simple problem.
I want to create a PDF with 20k pages.

If the PDF is with 10k node.js loads in memory the `Uint8Array` and the results from
`pdfDoc.save()` is saved with `this.writeStream.write();`

`const pdfBytes = await this.pdfDoc.save();`

await is needed since it get's only promise as result and not the Writer instance...

Looking in to the source of pdf-lib it seams to use stream for save()... 
According to: 
```
export interface SaveOptions {
  useObjectStreams?: boolean;
  addDefaultPage?: boolean;
  objectsPerTick?: number;
}
```

Still in the save method:

```
await this.flush();

const Writer = useObjectStreams ? PDFStreamWriter : PDFWriter;
return Writer.forContext(this.context, objectsPerTick).serializeToBuffer();
```

So the Uint8Array is recieved only after all 20k pages... are added.

Any suggestions how to go with this use case of adding pages and writing the file to theserver at the same time without making the heap out of memory ?

MikeMnD:
To prevent questions of type why do you need 10k pages in one pdf file: [Here a possible explanation](https://answers.acrobatusers.com/Maximum-number-of-pages-in-a-pdf-q215428.aspx)

Hopding:
Hello @MikeMnD!

`pdf-lib` is currently limited in its ability to modify (or create) extremely large documents. This is because, as you noted, everything is performed on the heap without any sort of I/O streaming (see #239). Thus, the size of documents that can be manipulated is limited by the amount of memory available to the JavaScript engine. 

That being said, there's no reason this couldn't be changed. It should be possible for `pdf-lib` to support I/O streaming. This would provide a number of obvious benefits, including support for creating/modifying documents of arbitrary size.  

However, as I mentioned in #239, there are a number of more pressing issues I need to address before I get around to this. But if anybody would be interested in implementing this themselves, I'd be more than happy to provide guidance and answer questions!

Hopding:
Added this to the roadmap for tracking: https://github.com/Hopding/pdf-lib/discussions/998.

wdelmas:
pdf-lib loads entire documents into memory with no streaming support. For your scenario consuming 6GB+:

**Why it happens:** Each page copies font subsets, embedded images get duplicated rather than referenced. The library wasn't designed for batch processing.

**Workarounds:**

[...]

PR #275, opened 2019-12-20
Acroform Field Base Class

### LINKED ISSUE #261: AcroFormField Structure
As part of adding general AcroForm functionality, we want a structure to represent an individual entry in the AcroForm Fields array 

### Tasks
- [x] Implement a `PDFAcroFormField` structure extending `PDFDict` following pg. 432-433 of the PDF spec
- [x] Update `PDFAcroForm` structure to return a `PDFArray` of `PDFAcroFormField` structures.

### Dependencies
- Requires https://github.com/Hopding/pdf-lib/pull/255

cshenks:
Fixes #261

I've added a `PDFAcroFormField` structure with methods to return the values in a form field dictionary. I've also added a lookup method to the context such that `PDFAcroForm.Fields` now returns an array of `PDFAcroFormField` instances


Hopding:
@cshenks This looks good at a high level. I'll review in detail once you have a chance to incorporate the changes from https://github.com/Hopding/pdf-lib/pull/255.

Also, I like the static members you added to `PDFName.ts`. They really help sharpen up the code.

cshenks:
@Hopding merged in the changes from `AcroFormsAPI`

src/core/structures/PDFAcroForm.ts
```diff
@@ -22,12 +23,15 @@ class PDFAcroForm {
...
-  fieldDicts(): PDFDict[] {
-    return this.Fields().lookupElements(PDFDict);
+  Fields(): PDFAcroFormField[] {
+    const fieldDicts = this.dict
+      .lookup(PDFName.of('Fields'), PDFArray)
+      .lookupElements(PDFDict);
+    const fields = [];
+    for (let idx = 0, size = fieldDicts.length; idx < size; idx++) {
+      fields.push(PDFAcroFormField.fromDict(fieldDicts[idx]));
+    }
```
Hopding: Since we're going through a loop with a known size, it's more efficient to allocate the array up front and assign to indexes, rather than pushing onto it. This improves performance, which is the only reason to use `for` loops like this to begin with. Also (and this is pedantic), I've tried to standardize the `for` loop variables to `idx, len`, rather than `idx, size`. The choice is arbitrary, but I do think that consistency is valuable.

For example:
```js
const fields = new Array(fieldDicts.length);
for (let idx = 0, len = fieldDicts.length; idx < len; idx++) {
  fields[idx] = PDFAcroFormField.fromDict(fieldDicts[idx]);
}
```

**Sidenote:** https://github.com/Hopding/pdf-lib/issues/197 is a great example of just how impressively large some PDF files can really be!

src/core/structures/PDFAcroFormField.ts
```diff
@@ -0,0 +1,75 @@
+import {
+  PDFArray,
+  PDFDict,
+  PDFName,
+  PDFNumber,
+  PDFObject,
+  PDFString,
+} from 'src/core';
+
+const acroFormFieldTypes: Array<PDFObject | undefined> = [
```
Hopding: TypeScript should be able to automatically infer the type of this array as `PDFRef[]`, so I don't think this type annotation is necessary at all. However, if it actually is necessary for some reason, we should refine it to `const acroFormFieldTypes: PDFRef[]`.

src/core/structures/PDFAcroFormField.ts
```diff
@@ -0,0 +1,75 @@
...
+  PDFName.Btn,
+  PDFName.Ch,
+  PDFName.Tx,
+  PDFName.Sig,
+];
+
+class PDFAcroFormField {
```
Hopding: What do you think about shortening this class name to `PDFAcroField`? I think it's just as clear, and the brevity makes it easier to read/use.

src/core/structures/PDFAcroFormField.ts
```diff
@@ -0,0 +1,75 @@
...
+    const ft = PDFName.of('FT');
+    const hasValidFieldType =
+      dict.has(ft) && acroFormFieldTypes.includes(dict.get(ft));
+    if (!hasValidFieldType) {
+      throw new Error('Invalid PDFAcroFormField Type');
+    }
+    return new PDFAcroFormField(dict);
+  }
+
+  private readonly dict: PDFDict;
```
Hopding: I think we should leave these underlying values `readonly`, but publicly accessible (same goes for [`PDFAcroForm.dict`](https://github.com/Hopding/pdf-lib/blob/00c8d7706c912bce9615afc087ed7b0afca7e95b/src/core/structures/PDFAcroForm.ts#L19), actually). In a perfect world, users would never need to dip into the low-level guts and could do everything they needed via the high level APIs. But in practice, there's always a minority of users that need to get into the nitty gritty of things.

src/core/objects/PDFDict.ts
```diff
@@ -45,6 +46,7 @@ class PDFDict extends PDFObject {
     return this.dict.has(key);
   }
 
+  lookupMaybe(key: PDFName, type: typeof PDFAcroForm): PDFAcroForm | undefined;
```
Hopding: Is this still necessary?

src/core/structures/PDFAcroFormField.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  protected constructor(dict: PDFDict) {
+    this.dict = dict;
+  }
+
+  FT(): PDFName {
```
Hopding: This field is actually only required for terminal fields. AcroFields can actually be structured as a tree. In which case, the branch nodes will not have a `FT`.

Speaking of which, it might be useful to distinguish between terminal an non-terminal fields. Because the `Kids` field is required for non-terminal fields, but is disallowed on terminal ones. We could structure it like so:

```js
class PDFAcroField // "dummy" base class, similar to `PDFObject`
class PDFTerminalField extends PDFAcroField
class PDFNonTerminalField extends PDFAcroField
```

But I'm open to alternatives as well. Or perhaps we should wait to build this abstraction until more work has been done on some of the other AcroForm API cards.

src/core/structures/PDFAcroForm.ts
cshenks: Updated

src/core/structures/PDFAcroFormField.ts
cshenks: Updated

src/core/structures/PDFAcroFormField.ts
cshenks: Updated

src/core/structures/PDFAcroFormField.ts
cshenks: Updated

src/core/objects/PDFDict.ts
cshenks: Updated

src/core/structures/PDFAcroFormField.ts
cshenks: Updated

PR #294, opened 2019-12-30
Acrotext additional properties

### LINKED ISSUE #293: Additional Acrofield Text Properties
As part of #262, let's add the additional, text-field-specific attributes detailed on pg. 443-444 of the PDF spec

cshenks:
Fixes #293 

Requires #300

### DIFF
--- a/src/core/structures/acroform/PDFAcroText.ts
+++ b/src/core/structures/acroform/PDFAcroText.ts
@@ -1,9 +1,45 @@
-import { PDFDict } from 'src/core';
+import { PDFDict, PDFName, PDFNumber, PDFObject } from 'src/core';
 import { PDFTerminalField } from './index';
 
 class PDFAcroText extends PDFTerminalField {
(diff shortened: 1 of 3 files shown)


src/core/structures/acroform/PDFAcroText.ts
```diff
@@ -1,10 +1,46 @@
...
+  RV(): PDFObject | undefined {
+    return this.dict.lookup(PDFName.RV);
+  }
+
+  MaxLen(): PDFNumber | undefined {
+    return this.dict.lookupMaybe(PDFName.MaxLen, PDFNumber);
+  }
+
+  isMultiLine(): boolean {
+    return !!(this.Ff().value() & (1 << 13));
```
Hopding: Should these be changed to use `this.getFlags()`?

cshenks: Updated

PR #306, opened 2020-01-04
Checkbox

### LINKED ISSUE #269: Check Boxes
### Description
As part of #267, let's distinguish check box fields from the generic `AcroFormFieldButton`

### Tasks
- [x] Create a `CheckBox` structure as described in Ch. 12.7.4.2.3 of the PDF spec
- [x] Cast all `Btn` fields with cleared pushbutton and radio flags to `CheckBox` in the `PDFAcroForm` `Fields` array

cshenks:
Fixes #269


src/core/structures/acroform/CheckBox.ts
```diff
@@ -0,0 +1,46 @@
+import { PDFArray, PDFDict, PDFName } from 'src/core';
+import { PDFAcroButton } from './index';
+
+class CheckBox extends PDFAcroButton {
```
Hopding: Can we name this class `PDFAcroCheckBox`? 

I also think we should rename `PushButton` to `PDFAcroPushButton` (I missed it in my review): 

https://github.com/Hopding/pdf-lib/blob/1ee2f4cd2329a21a9821e54f6a9b0a8b516d0d6b/src/core/structures/acroform/PushButton.ts#L4

cshenks: Updated

PR #312, opened 2020-01-11
Annotation dict

### LINKED ISSUE #309: Annotation Dictionaries
In order to support setting values on some acroform field types (e.g., radio buttons), the dictionaries referred to in a terminal field's `Kids` array must be dereferenced so the associated appearance dictionaries can be read. Let's add a `PDFAnnotation` structure to support this. See pg. 382-385 of the spec for details

### Tasks
- [x] Create `PDFAnnotation`

### Dependencies
- Requires #309

cshenks:
Fixes #309


src/core/structures/PDFRectangle.ts
```diff
@@ -0,0 +1,46 @@
...
+      normalized.set(0, this.array.lookup(2, PDFNumber));
+      normalized.set(2, this.array.lookup(0, PDFNumber));
+    }
+
+    if (normalized.lookup(1, PDFNumber) > normalized.lookup(3, PDFNumber)) {
+      normalized.set(1, this.array.lookup(3, PDFNumber));
+      normalized.set(3, this.array.lookup(1, PDFNumber));
+    }
+    return PDFRectangle.fromArray(normalized);
+  }
```
Hopding: 1. When do you see this operation being useful? I'm not sure I understand its utility.
2. The `normalized` array being created here is an orphan. It is not attached to its context in any way. So if somebody modifies it, the changes will not be reflected in the serialized document (unless they attach it themselves). I'm not sure what the best way to resolve this is. It _may_ not be an issue either. It really depends on how this method is intended to be used (which leads back to my first question).

src/core/structures/annotation/PDFAnnotation.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  Type(): PDFName | undefined {
+    return this.dict.lookupMaybe(PDFName.Type, PDFName);
+  }
+
+  Subtype(): PDFName {
+    return this.dict.lookup(PDFName.Subtype, PDFName);
+  }
+
+  Rect(): PDFRectangle {
```
Hopding: `PDFRectangle` is not a subclass of `PDFObject`, so returning it here violates our convention of only returning primitives that can be stored in the `PDFContext` from the low-level getters. This method should just return a `PDFArray`, and another helper could be created for returning a `PDFRectangle` structure.

src/core/structures/PDFRectangle.ts
cshenks: I could imagine a user wanting the AcroFormsAPI to let them create fields and manipulate their dimensions. If you think this is out of scope I can cut it

src/core/structures/annotation/PDFAnnotation.ts
cshenks: Updated

src/core/structures/PDFRectangle.ts
Hopding: @cshenks I agree that this will need to be possible. My concern here is that the newly created `PDFArray` is not actually attached to the context. I think the way to do this would be to define a method like the following:
```js
PDFAnnotation.setRectangle({
  x: number,
  y: number,
  width: number,
  height: number,
})
```

cshenks: @Hopding Broke this out into #320

PR #323, opened 2020-01-20
Radio buttons

### LINKED ISSUE #270: Radio Buttons
### Description
As part of #267, let's distinguish radio button fields from the generic `AcroFormFieldButton`

### Tasks
- [x] Create a `RadioButton` structure as described in Ch. 12.7.4.2.4 of the PDF spec
- [x] Cast all `Btn` fields with the radio button flag set to `RadioButton`s in the `PDFAcroForm` `Fields` array

### Dependencies 
- Requires #314

cshenks:
Fixes #270 

Requires #321


src/core/structures/acroform/PDFRadioButton.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  getValue(): PDFName {
+    const value = this.dict.lookup(PDFName.V);
+    if (!value) {
+      return PDFName.Off;
+    }
+    return this.dict.lookup(PDFName.V, PDFName);
+  }
+
+  _getChildOnAppearanceState(index: number): PDFName {
```
Hopding: Since we're using TypeScript, if these fields are supposed to be private, can we mark them using the `private` keyword instead of underscore prefixes?

cshenks: Updated

Issue #184, opened 2019-09-06
[Feature Request]: embedPdf, drawPdf


BoLaMN:
allowing of drawing multiple pdfs into 1 page like jpeg and pngs 

[draw-pdf.pdf](https://github.com/Hopding/pdf-lib/files/3584195/draw-pdf.pdf)

[https://github.com/galkahana/PDF-Writer/wiki/PDF-Embedding](https://github.com/galkahana/PDF-Writer/wiki/PDF-Embedding)

webdesq:
I'm also interested in such a feature; the current method using copyPages() from a donor PDF works, but does not allow a donor page to be 'drawn' over an existsing page in target document. In my case, I need to draw an advertorial page 'full scale' in my target page. The target page has different dimensions (but same ratio) as donor page.

Hopding:
Hello @BoLaMN and @webdesq! I agree that this would be a useful feature. It shouldn't be terribly difficult to implement either. Most of the necessary supporting code already exists in the pdf-lib codebase. It'd mostly be a matter of putting the pieces together.

If anybody would like to create a PR for this, I'd be happy to provide advice and answer questions! Otherwise I'll get around to this after fixing some of the outstanding bugs and higher priority issues.

4rlekin:
Hey, could anyone help me and point me to the relevant part of codebase for rendering this pdf's?
(Hopding mentioned that some supporting code for this is mostly done).
I'd be happy to try and implement this for the project Im working on and then backport it to lib with PR.

Hopding:
@4rlekin I'd suggest taking a look at #348 where @tessi has made some great progress on implementing this.

Hopding:
Version [`1.4.0`](https://www.npmjs.com/package/pdf-lib/v/1.4.0) is now published. It contains PDF embedding support thanks to @tessi! (see [this example](https://github.com/Hopding/pdf-lib/tree/b693c81a5f05f4aef5633abba4d6d5c71f7c9305#embed-pdf-pages)). The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.4.0).

You can install this new version with npm:

```
npm install pdf-lib@1.4.0
```

It's also available on unpkg:

- https://unpkg.com/pdf-lib@1.4.0/dist/pdf-lib.js
- https://unpkg.com/pdf-lib@1.4.0/dist/pdf-lib.min.js

As well as jsDelivr:

[...]

PR #348, opened 2020-02-07
Implement embedded PDFs

tessi:
Hello dear PDF-creating friends 👋 

This is my attempt to allow embedding of PDF pages in other pages.
The good news is that it does not crash, the bad news is that it does
not work :) This is my first time playing with PDF internals, this project and typescript. I am pretty confident in that there are many things to improve :)

It is an attempt to solve https://github.com/Hopding/pdf-lib/issues/184 (@4rlekin if you read this, feel free to help, or, as you prefer, continue thinking about an alternative implementation).

Architecture-wise it was my idea was to follow the implementation of
how we embed PNGs:
Have an `embedPdfDocument` method  on PdfDocument which uses a
`PdfPageEmbedder`.
On PdfPage the returned object can be inserted using
`drawEmbeddedPdfPage`.

The PDF page to be embedded is wrapped into an XObject (`8.10 FormXObjects` from the PDF spec). This involved some heavy wrestling against contexts and copied pages.

I adapted the scratchpad to illustrate how this could be used. Sadly, it
just renders an empty, white page. I guess I need help debugging this.

Things to improve:

* [x] Is the general idea sound or total nonsense? @Hopding 
* [x] I violated the `PDFImage` class by adding an `EmbeddedPdfPage` option to
  the existing image types. To me it sounds like a `PDFPage` is not an
image and should either not be supported by this class or we should
rename/rework this class. Normally the pdf embedder would not need `width` and `height`, but I needed to implement it because `PDFImage` wants it.
* [x] XObjects seem to support transformation matrices and transparency. This
  is great, but the normal PNG code scales/rotates their content
differently. We need to decide which is better.
* [x] My attempts to support better placement (rotation, skew, clipping, ...) is mostly commented out. I wanted to be able to insert *something* first, then make it better.
* [x] Add tests
* [x] Reset the scratchpad to have the original code before merging this to
  master
* [x] Update changelog and eventually other metadata

I would appreciate reviews (to improve my TypeScript as well as the architecture, usage) and would love help in finding out why this does not work yet.

### DIFF
--- a/src/api/EmbeddedPDFPage.ts
+++ b/src/api/EmbeddedPDFPage.ts
@@ -9,22 +9,23 @@ import { assertIs } from 'src/utils';
 export default class EmbeddedPDFPage implements Embeddable {
   /**
    * > **NOTE:** You probably don't want to call this method directly. Instead,
-   * > consider using the [[PDFDocument.embedPng]] and [[PDFDocument.embedJpg]]
-   * > methods, which will create instances of [[PDFImage]] for you.
+   * > consider using the [[PDFDocument.embedPdfDocument]] and
+   * > [[PDFDocument.embedPdfPage]] methods, which will create instances of
+   * > [[EmbeddedPDFPage]] for you.
    *
-   * Create an instance of [[PDFImage]] from an existing ref and embedder
+   * Create an instance of [[EmbeddedPDFPage]] from an existing ref and embedder
    *
-   * @param ref The unique reference for this image.
-   * @param doc The document to which the image will belong.
-   * @param embedder The embedder that will be used to embed the image.
+   * @param ref The unique reference for this embedded page.
+   * @param doc The document to which the embedded page will belong.
+   * @param embedder The embedder that will be used to embed the page.
    */
   static of = (ref: PDFRef, doc: PDFDocument, embedder: PDFPageEmbedder) =>
     new EmbeddedPDFPage(ref, doc, embedder);
 
-  /** The unique reference assigned to this image within the document. */
+  /** The unique reference assigned to this embedded page within the document. */
   readonly ref: PDFRef;
 
-  /** The document to which this image belongs. */
+  /** The document to which this embedded page belongs. */
   readonly doc: PDFDocument;
 
   private alreadyEmbedded = false;
--- a/src/api/EmbeddedPDFPage.ts
+++ b/src/api/EmbeddedPDFPage.ts
@@ -10,7 +10,7 @@ export default class EmbeddedPDFPage implements Embeddable {
   /**
    * > **NOTE:** You probably don't want to call this method directly. Instead,
    * > consider using the [[PDFDocument.embedPdfDocument]] and
-   * > [[PDFDocument.embedPdfPage]] methods, which will create instances of
+   * > [[PDFDocument.embedPage]] methods, which will create instances of
    * > [[EmbeddedPDFPage]] for you.
    *
    * Create an instance of [[EmbeddedPDFPage]] from an existing ref and embedder
--- a/src/api/EmbeddedPDFPage.ts
+++ b/src/api/EmbeddedPDFPage.ts
@@ -1,0 +1,62 @@
+import Embeddable from 'src/api/Embeddable';
+import PDFDocument from 'src/api/PDFDocument';
+import { PDFPageEmbedder, PDFRef } from 'src/core';
+import { assertIs } from 'src/utils';
+
+/**
+ * Represents a PDF page that has been embedded in a [[PDFDocument]].
+ */
+export default class EmbeddedPDFPage implements Embeddable {
+  /**
+   * > **NOTE:** You probably don't want to call this method directly. Instead,
+   * > consider using the [[PDFDocument.embedPng]] and [[PDFDocument.embedJpg]]
+   * > methods, which will create instances of [[PDFImage]] for you.
+   *
+   * Create an instance of [[PDFImage]] from an existing ref and embedder
+   *
+   * @param ref The unique reference for this image.
+   * @param doc The document to which the image will belong.
+   * @param embedder The embedder that will be used to embed the image.
+   */
+  static of = (ref: PDFRef, doc: PDFDocument, embedder: PDFPageEmbedder) =>
+    new EmbeddedPDFPage(ref, doc, embedder);
+
+  /** The unique reference assigned to this image within the document. */
+  readonly ref: PDFRef;
+
+  /** The document to which this image belongs. */
+  readonly doc: PDFDocument;
+
+  private alreadyEmbedded = false;
+  private readonly embedder: PDFPageEmbedder;
+
+  private constructor(
+    ref: PDFRef,
+    doc: PDFDocument,
+    embedder: PDFPageEmbedder,
+  ) {
+    assertIs(ref, 'ref', [[PDFRef, 'PDFRef']]);
+    assertIs(doc, 'doc', [[PDFDocument, 'PDFDocument']]);
+    assertIs(embedder, 'embedder', [[PDFPageEmbedder, 'PDFPageEmbedder']]);
+
+    this.ref = ref;
+    this.doc = doc;
+    this.embedder = embedder;
+  }
+
+  /**
+   * > **NOTE:** You probably don't need to call this method directly. The
+   * > [[PDFDocument.save]] and [[PDFDocument.saveAsBase64]] methods will
+   * > automatically ensure all embeddable pages get embedded.
+   *
+   * Embed this embeddable page in its document.
+   *
+   * @returns Resolves when the embedding is complete.
+   */
+  async embed(): Promise<void> {
+    if (!this.alreadyEmbedded) {
+      await this.embedder.embedIntoContext(this.doc.context, this.ref);
+      this.alreadyEmbedded = true;
+    }
+  }
+}
--- a/src/api/EmbeddedPDFPage.ts
+++ b/src/api/EmbeddedPDFPage.ts
@@ -28,6 +28,12 @@ export default class EmbeddedPDFPage implements Embeddable {
   /** The document to which this embedded page belongs. */
   readonly doc: PDFDocument;
 
+  /** The width of this page in pixels. */
+  readonly width: number;
+
+  /** The height of this page in pixels. */
+  readonly height: number;
+
   private alreadyEmbedded = false;
   private readonly embedder: PDFPageEmbedder;
 
@@ -42,7 +48,42 @@ export default class EmbeddedPDFPage implements Embeddable {
 
     this.ref = ref;
     this.doc = doc;
+    this.width = embedder.width();
+    this.height = embedder.height();
+
     this.embedder = embedder;
+  }
+
+  /**
+   * Compute the width and height of this page after being scaled by the
+   * given `factor`. For example:
+   * ```js
+   * embeddedPage.width  // => 500
+   * embeddedPage.height // => 250
+   *
+   * const scaled = embeddedPage.scale(0.5)
+   * scaled.width  // => 250
+   * scaled.height // => 125
+   * ```
+   * This operation is often useful before drawing a page with
+   * [[PDFPage.drawPage]] to compute the `width` and `height` options.
+   * @param factor The factor by which this page should be scaled.
+   * @returns The width and height of the page after being scaled.
+   */
+  scale(factor: number) {
+    assertIs(factor, 'factor', ['number']);
+    return { width: this.width * factor, height: this.height * factor };
+  }
+
+  /**
+   * Get the width and height of this page. For example:
+   * ```js
+   * const { width, height } = embeddedPage.size()
+   * ```
+   * @returns The width and height of the page.
+   */
+  size() {
+    return this.scale(1);
   }
 
   /**
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -19,6 +19,7 @@ import {
   PDFHexString,
   PDFName,
   PDFObjectCopier,
+  PDFPageEmbedder,
   PDFPageLeaf,
   PDFPageTree,
   PDFParser,
@@ -695,6 +696,24 @@ export default class PDFDocument {
     assertIs(png, 'png', ['string', Uint8Array, ArrayBuffer]);
     const bytes = toUint8Array(png);
     const embedder = await PngEmbedder.for(bytes);
+    const ref = this.context.nextRef();
+    const pdfImage = PDFImage.of(ref, this, embedder);
+    this.images.push(pdfImage);
+    return pdfImage;
+  }
+
+  async embedPdfDocument(document: PDFDocument): Promise<PDFImage> {
+    assertIs(document, 'document', [[PDFDocument, 'PDFDocument']]);
+    const embedder = await PDFPageEmbedder.forDocument(document);
+    const ref = this.context.nextRef();
+    const embeddedPdfPage = PDFImage.of(ref, this, embedder);
+    this.images.push(embeddedPdfPage);
+    return embeddedPdfPage;
+  }
+
+  async embedPdfPage(page: PDFPage): Promise<PDFImage> {
+    assertIs(page, 'page', [[PDFPage, 'PDFPage']]);
+    const embedder = await PDFPageEmbedder.forPage(page);
     const ref = this.context.nextRef();
     const pdfImage = PDFImage.of(ref, this, embedder);
     this.images.push(pdfImage);
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -716,25 +716,38 @@ export default class PDFDocument {
    * const pdfDoc = await PDFDocument.create();
    *
    * const sourcePdfUrl = 'https://pdf-lib.js.org/assets/with_large_page_count.pdf';
-   * const sourceBuffer = await fetch(sourcePdfUrl).then((res) => res.arrayBuffer());
-   * const sourcePdfDoc = await PDFDocument.load(sourceBuffer);
+   * const sourcePdf = await fetch(sourcePdfUrl).then((res) => res.arrayBuffer());
    *
    * // embed page 74 into pdfDoc
-   * const embeddedPage = await pdfDoc.embedPdfDocument(sourcePdfDoc,73);
+   * const [ embeddedPage ] = await pdfDoc.embedPdfDocument(sourcePdf, [ 73 ]);
    * ```
    *
-   * @param document The source PDF document that contains the page to be embedded
-   * @param pageIndex Optionally, the page index (starting at 0 for the first page) of the page to embed. Defaults to 0 (the first page).
-   * @returns Resolves with the embedded pdf page.
+   * @param pdf The input data containing a PDF document.
+   * @param indices The indices of the pages that should be embedded.
+   * @returns Resolves with an array of the embedded pages.
    */
   async embedPdfDocument(
-    document: PDFDocument,
-    pageIndex?: number,
-  ): Promise<EmbeddedPDFPage> {
-    assertIs(document, 'document', [[PDFDocument, 'PDFDocument']]);
-    const page = document.getPages()[pageIndex || 0];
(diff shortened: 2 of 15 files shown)


Hopding:
Hello @tessi!

Great work on this PR! You've done some impressive work here.

I was able to get this working after making a few changes to your `PDFPageEmbedder.embedIntoContext` method. I've copied the updated version of this method below:

```ts
import PDFDocument from 'src/api/PDFDocument';
import PDFPage from 'src/api/PDFPage';
import PDFRawStream from 'src/core/objects/PDFRawStream';
import PDFRef from 'src/core/objects/PDFRef';
import PDFStream from 'src/core/objects/PDFStream';
import PDFContext from 'src/core/PDFContext';
import PDFObjectCopier from 'src/core/PDFObjectCopier';
import { decodePDFRawStream } from 'src/core/streams/decode';
import PDFContentStream from 'src/core/structures/PDFContentStream';
import { mergeIntoTypedArray } from 'src/utils';

// ...

async embedIntoContext(context: PDFContext, ref?: PDFRef): Promise<PDFRef> {
  const { left, bottom, right, top } = this.boundingBox;
  const bbox = [left, bottom, right, top];

  const copier = PDFObjectCopier.for(this.page.doc.context, context);
  const copiedPage = copier.copy(this.page.node);
  const { Contents, Resources } = copiedPage.normalizedEntries();

  if (!Contents) throw new Error('Missing page.Contents!');

  const decodedContents: Uint8Array[] = new Array(Contents.size());
  for (let idx = 0, len = Contents.size(); idx < len; idx++) {
    const stream = Contents.lookup(idx, PDFStream);
    if (stream instanceof PDFRawStream) {
      decodedContents[idx] = decodePDFRawStream(stream).decode();
    } else if (stream instanceof PDFContentStream) {
      decodedContents[idx] = stream.getUnencodedContents();
    } else {
      throw new Error(`Unrecognized stream type: ${stream.constructor.name}`);
    }
  }

  // TODO: Should probably insert a newline in between each content array when
  // merging them, just to be safe
  const mergedContents = mergeIntoTypedArray(...decodedContents);

  const xObject = context.stream(mergedContents, {
    Type: 'XObject',
    Subtype: 'Form',
    FormType: 1,
    BBox: bbox,
    Matrix: this.transformationMatrix,
    // do we want to handle transparency?
    // https://github.com/galkahana/PDF-Writer/blob/edb7478dce955dfab61bbc2219a1ca7b85bed924/PDFWriter/DocumentContext.cpp#L992
    Resources,
  });

  if (ref) {
    context.assign(ref, xObject);
    return ref;
  } else {
    return context.register(xObject);
  }
}
```

Basically, the reason your code wasn't working is that you were copying an array of stream references as the content of the XObject stream. You can confirm that this is the case by saving your document without object streams: `pdfDoc.save({ useObjectStreams: false })` (note that this format is generally very useful for debugging documents during development as it makes the output PDF human readable). If you open the resulting document in a text editor and search for the definition of indirect object `5 0`, you'll find the following:

```
5 0 obj
<<
/Type /XObject
/Subtype /Form
/FormType 1
/BBox [ 0 0 595.28 841.89 ]
/Matrix [ 1 0 0 1 0 0 ]
/Resources <<
/Font <<
/Helvetica-7128 7 0 R
>>
/XObject <<
>>
>>
/Length 22
>>
stream
[ 9 0 R 8 0 R 10 0 R ]
endstream
endobj
```

Note the stream's content (the stuff between the `stream` and `endstream` keywords). It is an array of object references. This is because you grabbed the value of the source page's `Contents` entry and stuck it directly into the stream. Unfortunately, this won't work. The reason is that `Contents` is an array of stream references. But you want the XObject stream's content to be a sequence of content stream operators.

The solution I implemented in my refactored version of the  `PDFPageEmbedder.embedIntoContext` method is to lookup each stream and extract its contents. Once this is done, we can simply concatenate each content stream into a single stream of operators. And this single stream can be used as the content of the XObject stream. 

After making these changes, indirect object `5 0` looks like the following:

```
5 0 obj
<<
/Type /XObject
/Subtype /Form
/FormType 1
/BBox [ 0 0 595.28 841.89 ]
/Matrix [ 1 0 0 1 0 0 ]
/Resources <<
/Font <<
/Helvetica-7128 7 0 R
>>
/XObject <<
>>
>>
/Length 110
>>
stream
q
BT
0 0 0 rg
/Helvetica-7128 24 Tf
24 TL
1 0 0 1 5 25 Tm
<7064662D6C696220697320617765736F6D6521> Tj
T*
ET
Q
```

And the resulting document contains the copied page!

>

<kbd>

</kbd>

Hopding:
@tessi I should also mention that your implementation concept is perfectly sound. In fact, it's exactly what I was thinking of when I wrote up https://github.com/Hopding/pdf-lib/issues/184#issuecomment-569120029.

As far as the `PDFImage` goes, perhaps we should create a new class to represent embedded PDF objects (it could potentially represent things besides embedded pages too)? Maybe something like `PDFGraphic`? `PDFImage` could then be made to subclass `PDFGraphic`... But I haven't fully thought through this, so it might have some holes. It just came to my mind as a potential solution. Thoughts?

tessi:
@hopding thanks a torn for your feedback and even more for your refactored and working solution! You gave me a fish and helped me learn some fishing at the same time :)

Regarding PDFImage: I agree that extracting the embedded pdf page functionality into its own class is the way to go. I would say I'd like to decide later if and how that class connects to PDFImage.

My next step is to implement clipping/transformations on the embedded pdf. I'm not sure yet where exactly I should place the transformations: in the XObjects matrix or in the drawing code in PDFPage#drawEmveddedPdfPage. I will probably try both and stick with what's best.

Thanks again for your welcoming and extremely helpful comments. I'm hyped to continue working on this PR.

tessi:
Thanks to your help @Hopding. PDF embedding works, even with slightly more complex source PDFs which include images \o/ Also scaling, rotation and skewing works:

Screenshot 2020-02-09 at 14 02 02

For the implementation I decided to stick to the way the PNG/JPEG variants are implemented as I think its a more consistent user experience. Also, using transformation matrices on the XObject itself would probably (at least as far as I understand the spec) apply to all uses of the embedded PDF.

With my current solution it is still possible to apply a custom transformation matrix to the XObject (so we don't keep users from doing this). However, transforming during the `draw` step is much easier to use and should be the way we advertise in the documentation.

I implemented the TODO you left (insert newlines in between decoded streams) - it's probably not the most efficient code but seems to do the job.

In addition, I changed the draw options to use its own options object: `PDFPageDrawEmbeddedPdfPageOptions`. I renamed what used to be `width`/`height` in the image options to `xScale` and `yScale` since the underlying `scale` operation seems to scale relative (e.g. `0.5` for half the size) and not in absolute values (e.g. `200` to get 200 units).

I have not checked yet, but think that it's similar for PNGs/JPGs. Maybe we want to rename their `width` and `height` options too. At least to me `width` and `height` sounds like the user should give absolute numbers, but this may be because I'm not native English. What do you think?

Next, I will clean up the code (and rescue poor `PDFImage` from being misused) to make it properly reviewable. Then, I will add tests.

tessi:
> I have not checked yet, but think that it's similar for PNGs/JPGs. Maybe we want to rename their width and height options too. At least to me width and height sounds like the user should give absolute numbers, but this may be because I'm not native English. What do you think?

I just re-read the docs 🙈 `width`/`height` does in fact get relative sizes, but the code

```ts
   const jpgImage = await pdfDoc.embedJpg(jpgImageBytes)
   const jpgDims = jpgImage.scale(0.5)
   const page = pdfDoc.addPage()
   page.drawImage(jpgImage, {
     x: 25,
     y: 25,
     width: jpgDims.width,
     height: jpgDims.height,
     rotate: degrees(30)
   })
```

uses `width` and `height` from the `jpgDims` object. Still a little weird to me (why can't we directly give `0.5` into the `imageOptions`?). But it's not as weird as I first thought :)

tessi:
Another Update:

* The `PDFImage` class is restored to sanity again: Embeddable page stuff now lives in its own class `EmbeddablePDFPage`.
* I slightly refactored `PdfDocument.flush()` to be more DRY and added another interface `Embeddble` to `PDFImage`, `PDFFont`, and the new `EmbeddablePDFPage` to make sure they all implement `embed()`.
* Added documentation to all the things
* updated the scratchpad to showcase a more complex example of the new API - it is possible to embed a part of a PDF page, and rotate/place it however you want (note: I will restore the original scratchpad before merging)

|the soure page|only a part of the source page embedded and rotated|
|-|-|
|Screenshot 2020-02-09 at 23 42 54|Screenshot 2020-02-09 at 23 42 19|

Feature-wise I would say this PR is complete. We only miss some tests now. What do you think @Hopding ?

tessi:
I added tests for the parts that were tested before (-> the embedder). The relevant other parts (page, document) had no existing tests for jpges/pngs. So I thought I may get away by also not testing them?!

Next steps:

* [ ] get a review by @Hopding  and apply discussed changes
* [ ] reset the scratchpad
* [ ] prepare metadata. I found no CHANGELOG.md, is there other parts where I should document the changes?

Hopding:
@tessi Exciting progress! I'll review this as soon as I get a chance. I don't keep a changelog file. Instead I record the changes included in each release with GitHub's releases feature: https://github.com/Hopding/pdf-lib/releases. If you'd like to provide anything specific, just make a note in this thread and I'll include it in the release description.

One thing that we'll definitely need before merging this is some integration tests. If you're not familiar with them, here's a [quick overview](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-integration-tests) explaining how to run them. Please let me know if you have any questions about how to work with them. I think it would be good to update test1 to embed a document or two to test embedding pages in new documents. And then we'll want to add or update another one to test embedding pages in existing documents.

tessi:
Thanks @Hopding :)

Adding integration tests, indeed, is crucial :) I added some pdf-embedding to `test1` and `test2`. The pages now look like this:

### `test1.ts`

Here, I added a 4th page which embeds page 3 (rotated by 90 degrees and stretched to cover most of the page).
In addition, `normal.pdf` is overlayed on top.

Screenshot 2020-02-13 at 09 23 20

### `test2.ts`

Here, we add part of page 74 of the PDF spec (a figure about PDF encryption) to the bottom-left of every page.

Screenshot 2020-02-13 at 09 52 42

tessi:
> If you'd like to provide anything specific, just make a note in this thread and I'll include it in the release description.

I think I'm fine. Just wanted to take as much work off you for preparing a release (a fast release is in my interest, since I need the built package to use it in my project :))

src/api/PDFDocument.ts
```diff
@@ -701,27 +708,94 @@ export default class PDFDocument {
...
+   * ```
+   *
+   * @param document The source PDF document that contains the page to be embedded
+   * @param pageIndex Optionally, the page index (starting at 0 for the first page) of the page to embed. Defaults to 0 (the first page).
+   * @returns Resolves with the embedded pdf page.
+   */
+  async embedPdfDocument(
+    document: PDFDocument,
+    pageIndex?: number,
+  ): Promise<EmbeddedPDFPage> {
```
Hopding: I'd like to rename this method to `embedPdf`, just for the sake of brevity. I also think it would be best if this method accepted a raw PDF (similar to [`PDFDocument.load`](https://pdf-lib.js.org/docs/api/classes/pdfdocument#static-load)). These changes would keep the API more aligned with the existing `embed*` methods. And if the user wants to embed pages from an actual `PDFDocument` object, they can simply use the `PDFDocument.embedPdfPage` method.

Also, this API doesn't handle embedding multiple pages from a document very efficiently. Doing so would likely result in many duplicated objects, due to a new `PDFObjectCopier` being created and used for each page the user wants to copy. This inefficiency can be avoided by creating one `PDFObjectCopier` and using it to copy all of the desired pages for a given document. Of course, to achieve this, the APIs will need to be converted to accept and return arrays. For example, this method's signature would become:
```js
async embedPdf(
  pdf: string | Uint8Array | ArrayBuffer, 
  indices: number[],
): Promise<EmbeddedPDFPage[]>
```

I should note that this is not just a theoretical concern. The initial API I created for copying pages in version v0.x.x of pdf-lib did not handle copying multiple pages either. Users ran into the duplicate objects issue I explained above. So I changed the [page copying API](https://pdf-lib.js.org/docs/api/classes/pdfdocument#copypages) to accept an array of indices in v1.0.0 of pdf-lib.

src/api/PDFDocument.ts
```diff
@@ -701,27 +708,94 @@ export default class PDFDocument {
...
+   * @param page The source PDF page to be embedded
+   * @param boundingBox Optionally, area of the source page that should be embedded
+   * @param transformationMatrix Optionally, a transformation matrix that is always applied to the embedded page
+   * @returns Resolves with the embedded pdf page.
+   */
+  async embedPdfPage(
+    page: PDFPage,
+    boundingBox?: BoundingBox,
+    transformationMatrix?: TransformationMatrix,
+  ): Promise<EmbeddedPDFPage> {
```
Hopding: I'd like to rename this method to `embedPage`, just for the sake of brevity. Also, please see my comment above about accepting and returning arrays, as it applies to this method as well.

src/core/embedders/PDFPageEmbedder.ts
```diff
@@ -0,0 +1,153 @@
...
+    boundingBox?: BoundingBox,
+    transformationMatrix?: TransformationMatrix,
+  ) {
+    this.page = page;
+    this.boundingBox = boundingBox
+      ? boundingBox
+      : this.fullPageBoundingBox(page);
+    this.transformationMatrix = transformationMatrix
+      ? transformationMatrix
+      : identityMatrix;
```
Hopding: This can be simplified a bit:
```js
this.boundingBox = boundingBox || this.fullPageBoundingBox(page);
this.transformationMatrix = transformationMatrix || identityMatrix;
```

src/core/embedders/PDFPageEmbedder.ts
```diff
@@ -0,0 +1,153 @@
...
+      context.assign(ref, xObject);
+      return ref;
+    } else {
+      return context.register(xObject);
+    }
+  }
+
+  // `contents` is an array of streams which are merged to include them in the XObject.
+  // This methods extracts each stream and joins them with a newline character.
+  private decodeContents(contents: PDFArray) {
```
Hopding: I think we can simplify the newline logic of this method a bit:
```js
import CharCodes from 'src/core/syntax/CharCodes';

private decodeContents(contents: PDFArray) {
  const newline = Uint8Array.of(CharCodes.Newline);
  const decodedContents: Uint8Array[] = [];

  for (let idx = 0, len = contents.size(); idx < len; idx++) {
    const stream = contents.lookup(idx, PDFStream);

    let content: Uint8Array;
    if (stream instanceof PDFRawStream) { ... } 
    else if (stream instanceof PDFContentStream) { ... } 
    else { throw ... }

    decodedContents.push(content, newline)
  }

  return mergeIntoTypedArray(...decodedContents);
}
```

src/api/PDFPage.ts
```diff
@@ -678,6 +681,65 @@ export default class PDFPage {
...
+   *   y: 200,
+   *   xScale: 0.5,
+   *   yScale: 0.5,
+   *   rotate: degrees(30)
+   * });
+   * ```
+   * @param embeddedPdfPage The embeddedPDF page object to be drawn.
+   * @param options The options to be used when drawing the pdf page.
+   */
+  drawEmbeddedPdfPage(
```
Hopding: Can we rename this to `drawPage`, for the sake of brevity? I think the docs and method signature itself are sufficient to make it clear that the method handles `EmbeddedPDFPage` objects.

src/core/embedders/PDFPageEmbedder.ts
```diff
@@ -0,0 +1,153 @@
...
+      ? boundingBox
+      : this.fullPageBoundingBox(page);
+    this.transformationMatrix = transformationMatrix
+      ? transformationMatrix
+      : identityMatrix;
+  }
+
+  async embedIntoContext(context: PDFContext, ref?: PDFRef): Promise<PDFRef> {
+    const copier = PDFObjectCopier.for(this.page.doc.context, context);
+    const copiedPage = copier.copy(this.page.node);
```
Hopding: We should check to see if the page's context is different from the specified `context` argument, and only create a copy if it is. If it's the same, there's no need to copy anything. Doing so will only create duplicate objects and bloat the size of the resulting PDF file. For example:
```js
const { page } = this;
const copier = PDFObjectCopier.for(page.doc.context, context);
const copiedPage = context === page.doc.context ? page : copier.copy(page.node);
```

src/api/EmbeddedPDFPage.ts
```diff
@@ -0,0 +1,62 @@
...
+   * > **NOTE:** You probably don't want to call this method directly. Instead,
+   * > consider using the [[PDFDocument.embedPng]] and [[PDFDocument.embedJpg]]
+   * > methods, which will create instances of [[PDFImage]] for you.
+   *
+   * Create an instance of [[PDFImage]] from an existing ref and embedder
+   *
+   * @param ref The unique reference for this image.
+   * @param doc The document to which the image will belong.
+   * @param embedder The embedder that will be used to embed the image.
+   */
```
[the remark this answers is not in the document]
tessi: oh my. thanks for finding these copy&paste errors :)

src/core/embedders/PDFPageEmbedder.ts
tessi: good point! will implement this (and probably extract the copying into a function)

Hopding:
@tessi Just wanted to check and see if you've had a chance to read over and consider all my comments?

tessi:
Hey @Hopding I've seen your comments and as always they seem reasonable, well argued and nice :) I started working on them, but then came vacations and some family time. But I'm on it 🤞 (some of the easier changes are already commited and pushed)

Hopding:
Sounds good! Looking forward to merging this 👍

ihmc3jn09hk:
When will this PR be merged ?
This PR will help on a work around for #363, #202. Say svg-to-pdfkit -> pdfkit-to-pdf_lib -> embedPDF.

src/api/PDFDocument.ts
tessi: I agrees, done! :)

tessi:
@ihmc3jn09hk I guess this will be merged when I find enough free time to work on this (which I am motivated to do, but without any promises) and Hopding finds enough free time to review again.

I guess we cannot promise a specific time

tessi:
To my own surprise, I found some extended time to work on this PR today. @Hopding do you mind having another look? Changes are:

* implemented width/height/scale for the embedded PDF pages
* applied the renames you wished for
* made `embedDocument` multi-page aware
* fixed typos and copy&paste error you outlined
* applied the code improvements you suggested
* adapted the integration tests to use both, `xScale`/`yScale` and `width`/`height` when drawing pages
* extended specs for `PDFPageEmbedder` to test that it uses not only the page dimensions, but those from the given boundingBox

tessi:
@Hopding thanks for merging, and even more for your continuous support in that journey. It was fun working with you 🍻

Hopding:
@tessi It was great working with you as well! I really appreciate the work you did on this feature. I'm confident it will be a popular one.

Hopding:
Version [`1.4.0`](https://www.npmjs.com/package/pdf-lib/v/1.4.0) is now published. It contains PDF embedding support. (see [this example](https://github.com/Hopding/pdf-lib/tree/b693c81a5f05f4aef5633abba4d6d5c71f7c9305#embed-pdf-pages)). The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.4.0).

```
npm install pdf-lib@1.4.0
```

It's also available on unpkg:

- https://unpkg.com/pdf-lib@1.4.0/dist/pdf-lib.js
- https://unpkg.com/pdf-lib@1.4.0/dist/pdf-lib.min.js

As well as jsDelivr:

- https://cdn.jsdelivr.net/npm/pdf-lib@1.4.0/dist/pdf-lib.js
- https://cdn.jsdelivr.net/npm/pdf-lib@1.4.0/dist/pdf-lib.min.js

Issue #210, opened 2019-10-03
[Feature Request]: `opacity` option for `drawText()` method


akaanish:
I want to draw text with opacity 50% but i can see i cannot pass alpha value in rgb()

function.

```
export const rgb = (red: number, green: number, blue: number): RGB => {
  assertRange(red, 'red', 0, 1);
  assertRange(green, 'green', 0, 1);
  assertRange(blue, 'blue', 0, 1);
  return { type: ColorTypes.RGB, red, green, blue };
};

 drawText(text: string, options: PDFPageDrawTextOptions = {}): void {
    assertIs(text, 'text', ['string']);
    assertOrUndefined(options.color, 'options.color', [[Object, 'Color']]);
    assertOrUndefined(options.font, 'options.font', [[PDFFont, 'PDFFont']]);
    assertOrUndefined(options.size, 'options.size', ['number']);
    assertOrUndefined(options.rotate, 'options.rotate', [[Object, 'Rotation']]);
    assertOrUndefined(options.xSkew, 'options.xSkew', [[Object, 'Rotation']]);
    assertOrUndefined(options.ySkew, 'options.ySkew', [[Object, 'Rotation']]);
    assertOrUndefined(options.x, 'options.x', ['number']);
    assertOrUndefined(options.y, 'options.y', ['number']);
    assertOrUndefined(options.lineHeight, 'options.lineHeight', ['number']);
    assertOrUndefined(options.maxWidth, 'options.maxWidth', ['number']);
    assertOrUndefined(options.wordBreaks, 'options.wordBreaks', [Array]);

    const [originalFont] = this.getFont();
    if (options.font) this.setFont(options.font);
    const [font, fontKey] = this.getFont();

    const fontSize = options.size || this.fontSize;

    const wordBreaks = options.wordBreaks || this.doc.defaultWordBreaks;
    const textWidth = (t: string) => font.widthOfTextAtSize(t, fontSize);
    const lines =
      options.maxWidth === undefined
        ? cleanText(text).split(/[\r\n\f]/)
        : breakTextIntoLines(text, wordBreaks, options.maxWidth, textWidth);

    const encodedLines = new Array(lines.length) as PDFHexString[];
    for (let idx = 0, len = lines.length; idx < len; idx++) {
      encodedLines[idx] = font.encodeText(lines[idx]);
    }
```
I cannot see how i can set opacity value.
Does this library support?
If not, it would be really nice if you consider it.

Hopding:
[...]

PR #487, opened 2020-06-15
Opacity - semi transparent objects

soadzoor:
Hey @Hopding ,

So I thought I'd submit a PR for #210 , as you suggested yesterday.

Some notes:
- I've tested it, it works for my case, but there might be some cases I've missed.
- Maybe I shouldn't push a new graphic state if both `opacity` and `borderOpacity` is 1 (or left untouched, as the default value is 1 for both of them)? I'm not sure about best practices when it comes to PDF.
- Since it's only for `svgPath`s so far, I'd still need to add these things to `drawText` (as in the original issue), and a lot of other stuff (`drawLine`, `drawCircle`, `drawSquare`, `drawImage`, etc). In that case, I should try to avoid most of the code duplication related to this feature with private methods for example. Please let me know if you have any preferences, thoughts, warnings about this!

Have a great week!

### DIFF
--- a/apps/deno/tests/test12.ts
+++ b/apps/deno/tests/test12.ts
@@ -93,6 +93,22 @@ export default async (_assets: Assets) => {
       borderWidth: 2,
     },
   );
+  page2.drawCircle({
+    x: inchToPt(3),
+    y: inchToPt(5),
+    color: rgb(0, 1, 1),
+    opacity: 0.1,
+    borderWidth: 3,
+    borderColor: rgb(1, 0, 1),
+    borderOpacity: 0.2
+  });
+  page2.drawText("Semi-Transparent Text", {
+    color: rgb(0, 1, 1),
+    opacity: 0.5,
+    x: inchToPt(1),
+    y: inchToPt(2.5),
+    size: 50
+  });
 
   const pdfBytes = await pdfDoc.save();
   return pdfBytes;
--- a/apps/deno/tests/test2.ts
+++ b/apps/deno/tests/test2.ts
@@ -74,6 +74,7 @@ export default async (assets: Assets) => {
       width: textWidth + 10,
       height: boxHeight,
       color: solarizedWhite,
+      opacity: 0.85,
       borderColor: solarizedGray,
       borderWidth: 3,
     });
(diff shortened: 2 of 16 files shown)


soadzoor:
... I wanted to create a draft pull request, but I misclicked. Now I can't see an option to change this to a draft PR.. Any ideas? :(

soadzoor:
@Hopding I've updated this PR, so now it contains opacity (and borderOpacity) for all the draw options (text, image, page, rectangle, line, square, ellipse, circle, svgpath)

Hopding:
@soadzoor Thanks for working on this! I'll take a look. No worries about the draft PR. I don't think the status can be changed once it's opened, but that's no big deal.

Hopding:
@soadzoor Can you please update a few of the integration tests to use this feature?

soadzoor:
@Hopding Sure, if you tell me how to do that! :)

Hopding:
@soadzoor Have you been able to run the integration tests? https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-integration-tests

soadzoor:
@Hopding I haven't checked that one yet! It's past midnight here, so I'll do it tomorrow if that's okay :)

Hopding:
Sounds good! No rush

soadzoor:
@Hopding I've updated test 2, 4, and 12, so now they contain some kind of opacity. I manually copy-pasted the same modifications from html to node/ts, deno/ts, and rn/js, and I'm not sure if that's the proper way of doing it. I didn't feel very professional while I was doing it, but for tests, it might be okay, I don't know. Please let me know what you think!

Hopding:
@soadzoor The tests look good! Thanks!

And yes, there's a lot of copy/pasting between the tests for Node, Deno, React Native, and the Browser. It does feel a bit clunky, but it is by far the simplest approach and isn't a big deal since they're just tests. In a previous version of pdf-lib I tried to avoid the copy/pasting and reuse the tests. It turned into a complicated mess because each environment needs things to be a little bit different.

soadzoor:
@Hopding I see your point, it's completely fine for tests, I think. :)
I don't want to sound pushy, and I know you still need to review the code, but when do you think this PR will be merged? Just a rough estimation is okay!

Let me know if you'd need any modifications to this!

Hopding:
I'll probably merge it this evening. But I might make a few tweaks and also need to run through all the tests before cutting a release. So I'd expect to cut a release with this feature sometime this weekend.

soadzoor:
@Hopding Sounds good! Thank you!

PR #498, opened 2020-06-21
Added highlevel ability to provide dashArray and dashPhase

soadzoor:
Hey @Hopding ,

To make this awesome lib even more awesome, and easier to use, I've create this PR.
This PR provides ability to simply pass the dashArray and dashPhase as parameters to the following page.draw functions:
- circle,
- ellipse,
- line,
- rectangle,
- square,
- svgPath

Please, let me know if needs any modifications.

Cheers!

### DIFF
--- a/apps/deno/tests/test1.ts
+++ b/apps/deno/tests/test1.ts
@@ -155,6 +155,20 @@ export default async (assets: Assets) => {
   });
   page1.pushOperators(popGraphicsState());
 
+  page1.drawLine({
+    start: {
+      x: size / 4,
+      y: size / 4
+    },
+    end: {
+      x: size / 4 + 100,
+      y: size / 4 + 100
+    },
+    color: rgb(0, 1, 0),
+    thickness: 3,
+    dashArray: [12, 12]
+  });
+
   // Lower-right quadrant
   page1.moveTo(size / 2, 0);
   page1.drawSquare({ size: size / 2, color: grayscale(0.8) });
--- a/apps/deno/tests/test1.ts
+++ b/apps/deno/tests/test1.ts
@@ -98,8 +98,6 @@ export default async (assets: Assets) => {
       xSkew: degrees(0),
       ySkew: degrees(0),
       color: undefined,
-      dashArray: [],
-      dashPhase: 0,
     }),
     popGraphicsState(),
   );
@@ -114,6 +112,9 @@ export default async (assets: Assets) => {
     xScale: 25,
     yScale: 150,
     color: rgb(255 / 255, 153 / 255, 51 / 255),
+    borderWidth: 2,
+    borderColor: rgb(0, 1, 1),
+    borderDashArray: [10],
   });
   page1.drawEllipse({
     x: size / 2 + size / 4,
@@ -152,8 +153,8 @@ export default async (assets: Assets) => {
     y: size / 4,
     size: 150,
     borderWidth: 10,
-    dashArray: [25],
-    dashPhase: 25,
+    borderDashArray: [25],
+    borderDashPhase: 25,
     borderColor: cmyk(0, 1, 0, 0),
   });
   page1.pushOperators(popGraphicsState());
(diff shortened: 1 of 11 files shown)


src/api/PDFPageOptions.ts
```diff
@@ -51,15 +51,19 @@ export interface PDFPageDrawSVGOptions {
   opacity?: number;
   borderColor?: Color;
   borderOpacity?: number;
+  dashArray?: number[];
+  dashPhase?: number;
```
Hopding, line 55: Can we call these options `borderDashArray` and `borderDashPhase`? This maintains the convention of prefixing all options related to the border with `border...`. And, of course, this will also need to be changed most of the other places these options are used.

src/api/PDFPageOptions.ts
```diff
@@ -51,15 +51,19 @@ export interface PDFPageDrawSVGOptions {
...
 export interface PDFPageDrawLineOptions {
   start: { x: number; y: number };
   end: { x: number; y: number };
   thickness?: number;
   color?: Color;
-  lineCap?: LineCapStyle;
   opacity?: number;
+  lineCap?: LineCapStyle;
+  dashArray?: number[];
+  dashPhase?: number;
```
Hopding, line 66: These **should** continue to be called `dashArray` and `dashPhase` (without the `border...` prefix), because they apply to the line itself, not its border.

src/api/operations.ts
```diff
@@ -151,13 +152,16 @@ export const drawLine = (options: {
   thickness: number | PDFNumber;
   color: Color | undefined;
   lineCap?: LineCapStyle;
+  dashArray: (number | PDFNumber)[];
+  dashPhase: number | PDFNumber;
```
Hopding, line 156: Adding required options to this interface is actually a breaking change (I know, because I've made the mistake before 😂). So these options need to be marked as options, just like `graphicsState` is. For example:
```ts
  dashArray?: (number | PDFNumber)[];
  dashPhase?: number | PDFNumber;
```
And then we'll need to check if they're present when applying the `setDashPattern` operator:
```ts
setDashPattern(options.dashArray ?? [], options.dashPhase ?? 0)
```

src/api/PDFPage.ts
```diff
@@ -1143,6 +1145,8 @@ export default class PDFPage {
         color: options.color ?? undefined,
         borderColor: options.borderColor ?? undefined,
         borderWidth: options.borderWidth ?? 0,
+        dashArray: options.dashArray || [],
+        dashPhase: options.dashPhase ?? 0,
```
Hopding, line 1149: Per my above comments, these two options will need to be made optional. I think that once we do that, I think it would be best to remove the `|| []` and `?? 0` defaults.

apps/deno/tests/test1.ts
```diff
@@ -98,6 +98,8 @@ export default async (assets: Assets) => {
       xSkew: degrees(0),
```
Hopding, line 98: These are good integration test updates. However, it's important that we test all page drawing methods that the dash array and dash phase options can be applied to. Right now we're only testing `PDFPage.drawCircle` and `PDFPage.drawLine`.

src/api/PDFPageOptions.ts
soadzoor: You're right, done!

src/api/PDFPage.ts
soadzoor: Done, and added it like this: `borderDashArray: options.borderDashArray ?? undefined`, because this is how the rest of the optional parameters are handled. Although I'm not sure about the purpose of `?? undefined`, can you explain why that is needed?

apps/deno/tests/test1.ts
soadzoor: I've just added drawRectangle, drawEllipse, and drawSquare with (border)DashArray in addition. And we already had one for drawSvgPath within test12. I think we cover all cases now

src/api/PDFPage.ts
Hopding: It shouldn't be necessary. I'm not actually sure how that wound in there. Maybe just the accidental result of some refactoring.

soadzoor:
@Hopding Awesome! Thank you!

src/api/PDFPage.ts
Hopding: No need to worry about removing them for now. I (or somebody else) will take them out later. ;)

PR #499, opened 2020-06-21
Added highlevel ability to provide lineCap style

soadzoor:
Similar to #498 , but for lineCap.

There was a comment from @Hopding , saying:
// TODO: Should assert `options.lineCap` here, but is a breaking change

I'm not sure why this would be a breaking change, we can just default to `LineCapStyle.Butt`, or we can ignore this parameter if it's not provided (like this PR does). Maybe there's more to it than I can see, so let me know! :)

### DIFF
--- a/apps/deno/tests/test1.ts
+++ b/apps/deno/tests/test1.ts
@@ -155,6 +155,20 @@ export default async (assets: Assets) => {
   });
   page1.pushOperators(popGraphicsState());
 
+  page1.drawLine({
+    start: {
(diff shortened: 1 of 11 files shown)


soadzoor:
@Hopding All done, let me know if it needs anything else! :)

Hopding:
@soadzoor Could you please merge `master` into your branch and resolve these conflicts?

soadzoor:
@Hopding sure, done!

Issue #473, opened 2020-06-05
[BUG] embedding image with alpha channel broken


AmilKey:
Hello everyone, I use embedding images with transparency to my document, but this function was broken in version 1.3.1, I have attached examples of generated documents in different versions.

example embedding
https://jsfiddle.net/d1uLne7g/

looks like that was broken into version 1.3.2
https://github.com/Hopding/pdf-lib/compare/v1.3.1..v1.3.2#diff-4f449742b7e9f5f136fe91da08ad248d

attached 2 pdf generated in version 1.3.1 and 1.7.0
[1.3.1.pdf](https://github.com/Hopding/pdf-lib/files/4736755/1.3.1.pdf)
[1.7.0.pdf](https://github.com/Hopding/pdf-lib/files/4736756/1.7.0.pdf)

@Hopding

Hopding:
Version [`1.9.0`](https://www.npmjs.com/package/pdf-lib/v/1.9.0) is now published. It contains the fix for this issue. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.9.0).

You can install this new version with npm:

```
npm install pdf-lib@1.9.0
```

It's also available on unpkg:

- https://unpkg.com/pdf-lib@1.9.0/dist/pdf-lib.js
- https://unpkg.com/pdf-lib@1.9.0/dist/pdf-lib.min.js

As well as jsDelivr:

- https://cdn.jsdelivr.net/npm/pdf-lib@1.9.0/dist/pdf-lib.js
- https://cdn.jsdelivr.net/npm/pdf-lib@1.9.0/dist/pdf-lib.min.js

PR #502, opened 2020-06-23
Fix for #483 - Semi-transparent PNG rendered opaque in PDF

### LINKED ISSUE #483: Semi-transparent PNG rendered opaque in PDF
Im trying to watermark a PDF with a semi-transparent PNG, but the semi-transparent portions are being rendered fully opaque in resulting PDF.

I can't find anything in the docs regarding transparency... any help would be greatly appreciated.

soadzoor:
Fixes #483 and #473

### DIFF
--- a/src/utils/png.ts
+++ b/src/utils/png.ts
@@ -58,7 +58,7 @@ export class PNG {
 
     this.rgbChannel = rgbChannel;
 
-    const hasAlphaValues = alphaChannel.some((a) => a < 1);
+    const hasAlphaValues = alphaChannel.some((a) => a < 255);
     if (hasAlphaValues) this.alphaChannel = alphaChannel;
 
     this.type = getImageType(upng.ctype);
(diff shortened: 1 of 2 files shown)


src/utils/png.ts
```diff
@@ -58,7 +58,7 @@ export class PNG {
 
     this.rgbChannel = rgbChannel;
 
-    const hasAlphaValues = alphaChannel.some((a) => a < 1);
+    const hasAlphaValues = alphaChannel.some((a) => a < 255);
```
Hopding, line 61: Thanks for tracking this down and fixing it!!! Much appreciated. Would you mind adding a quick unit test to prevent regressions? This will require a new test file: `tests/utils/png.spec.ts`.

soadzoor: @Hopding done! Let me know if it's okay (it's been a while since I wrote unit tests). That long line of Uin8Array might not be okay, but I'm not sure how to format that. It's ugly either way.. 😆 

The test itself passes with this PR (a < 255), and it fails without it (a < 1), so it's doing its job.

Hopding: Thanks @soadzoor. Just looked over it briefly and it looks good. I’ll try to get it merged this evening and released as soon as possible.

PR #503, opened 2020-06-23
Add blendMode options

bockoblur:
Added `blendMode` optional property for drawing methods.
Added tests for Node and Deno. (`test14.ts`)


apps/deno/index.ts
```diff
@@ -150,7 +151,7 @@ const main = async () => {
   // prettier-ignore
   const allTests = [
       test1, test2, test3, test4, test5, test6, test7, test8, test9, test10,
-      test11, test12, test13
+      test11, test12, test13, test14
```
Hopding, line 154: Instead of creating a new test, can we update some of the drawing operations for existing tests (or add a page if necessary)? The integration tests cannot easily be automated and they take a lot of time to run on all the platforms prior to each release, so I try to keep the total number of tests to a minimum. I prefer to exercise as much functionality in as few tests and pages as possible.

apps/deno/tests/test14.ts
```diff
@@ -0,0 +1,71 @@
+import { Assets } from '../index.ts';
+import { PageSizes, PDFDocument, BlendMode, degrees, rgb } from '../../../dist/pdf-lib.esm.js';
+
+
+const modeNames = Object.values<string>(BlendMode);
+const objectSize = 80;
+const numObjects = 60;
+
+export default async (_assets: Assets) => {
```
Hopding, line 9: This is a nice test (I enjoy the output). But please see my previous comment about testing the `blendMode` option in the drawing operations of existing tests. Also, it's important to test the `blendMode` option on every draw method that supports it. We're currently only testing the `drawCircle`, `drawSquare`, and `drawImage` operations.

bockoblur:
Hello, @Hopding. 

This push  should do it.

- deleted unnecessary files (`.vscode/settings.json` & `apps/deno/tests/blendmodes.pdf`)
- ran linter
- appended new tests to `test12.ts` (adds new page at the end) and deleted `test14.ts` (Both Node & Deno)
  - test includes `drawRectangle`, `drawText`, `drawSvgPath`, `drawCircle`, `drawImage` and `drawPage`
- added the same changes to `test12.html` for web test

I did not altered `rn` test, since I'm not fluent in React Native, so I didn't want to break something. :)

I hope you find it useful.

Issue #504, opened 2020-06-26
[Feature Request] API to get image resolution


bockoblur:
When placing images to PDF, it would be very useful to also know image **resolution**, in other words its **physical size**.

`PDFImage` methods `.size` and `.scale` return image size **in pixels**, and we don't have a way of finding the intended/physical image size **in mm/Inch/pt**, which would be useful if we want to place the image to page at its original, intended/physical size.

Here is an hypothetical example:

```TypeScript
  let imgBytes = ...
  let pdfDoc = await PDFDocument.create();
  let page = pdfDoc.addPage();
  const jpgImage = await pdfDoc.embedJpg(imgBytes);

  // **HYPOTHETICAL METHOD!**
  const resolution = jpgImage.resolution();

  // then we could calculate image scale factor to match the original image size:
  const jpgDims = jpgImage.scale( 72 / resolution);

  // or we could calculate image dimensions in units other than pixels
  const widthInInches = jpgDims.width / resolution
```

For JPG images, this could probably be done in `src/core/embedders/JpegEmbedder.ts` by looking in the image header, namely `Units` and `Xdensity & Ydensity`, but new property `resolution` would have to be added to the class.

For PNG, `UPNG` library used can be tapped into, reading `pHYs` data chunk which (if present) contains the physical measures of PNG file.

@Hopding if you are interested, I could PR and try to tackle this.

bockoblur:
Actually, JPG images are much tougher to parse, because there are (at least) two standards/containers: `JFIF` and `EXIF`, and
`EXIF` is very convoluted. Since I believe that adding another dependency just for this is not an option, I will try to write simplified parser for `EXIF`, to get image resolution/density.

 `PNG` and `JFIF` jpegs are rather straightforward and I have a working code prepared.

Hopding:
@bockoblur This feature makes sense to me. I'd be happy to accept a PR implementing this!

A couple of things I want to note (that you may have already thought of):
* Parsing is error prone. So if "extra" parsing has to be done on an image (e.g. parsing the `EXIF` metadata) to obtain it's resolution, that parsing should be done lazily. I'd imagine it happens only when `resolution()` is first called for that image and the values are then stored in a [`Cache`](https://github.com/Hopding/pdf-lib/blob/master/src/utils/Cache.ts).
[...]

PR #509, opened 2020-06-30
Get embedded image resolution

bockoblur:
Addressing #504 Get embedded image resolution

A few notes:
  - I opted for using dependency [ts-exif-parser](https://github.com/bpatrik/ts-exif-parser) for Exif parsing. 
  - also, as explained in my [comment](https://github.com/Hopding/pdf-lib/issues/504#issuecomment-650808177), I am sticking to returning a `number`, and not `number | undefined`, with fallback to 72 dpi resolution when no information is found in the file.
  - in [another comment](https://github.com/Hopding/pdf-lib/issues/504#issuecomment-650810039), I showed some measurements and found that parsing images for info about resolution is not very expensive, so for now I did not implement lazy image parsing. Also, if we can have `undefined` resolution, that would cause Cache to parse the image on every call, since `undefined` is also signal that there's no cached value. 
  - I made only `node` integration test for now (changed `test1.ts`, adding new page to resulting pdf) and added one more image to `assets` (png file with resolution info in it). I can do more tests when we polish what I did so far.

Cheers.


Hopding:
@bockoblur I generally try to avoid introducing new dependencies as they can turn into maintenance headaches. Of course, the same is true of trying to write everything yourself when robust dependencies already exist. I think in this case it makes sense to add a dependency for parsing EXIF metadata. 

But I want to be careful about which dependencies we add. `ts-exif-parser` makes me nervous. It has just 9 stars on GitHub and 454 downloads/week on NPM. It also has a transitive dependency on `sax` (though that module looks much more robust, with 16,487,083 downloads/week on NPM).

The `exifr` module (https://github.com/MikeKovarik/exifr) looks more robust with 321 GitHub stars and 20,000 downloads/week on NPM. Have you considered using this module instead? (I'm assuming it would work just as well - but I haven't tried it myself).

Also, your observations about the performance of the additional parsing are appreciated. It's good to know it doesn't introduce much additional overhead. However, I'm worried about the potential for parsing errors. It would be bad if somebody is able to successfully embed and draw a JPG image with the current version of pdf-lib, but we release an update that causes the embedding to fail because the EXIF parsing is unsuccessful. So we can cache the value or not - I don't mind too much either way. But I do want to make sure no new parsing occurs unless the user explicitly asks for it.

bockoblur:
@Hopding 
I just pushed [new commit](https://github.com/Hopding/pdf-lib/commit/a52dadf88c2b29da861cfcc36042cc7fc295178c)

- using `exifr` as dependency
- lazy parsing for resolution
- new method `.resolution()` **always** succeeds. If image is invalid or no reolution info is found in the image, default resolution (72dpi) is returned. 

Note: since `parser` method in `exifr` library is `async`, so is our new `.resolution()` method. This was not necessary for png files, but I made that one `async `too for consistency . Image is not parsed until first call to `.resolution()` method.

Please review and if feedback is positive I can write some more tests (for Deno & browser).

Hopding:
@bockoblur I think your approach makes sense. Although I just thought of another way we could prevent parsing errors and avoid some of the async methods and lazy parsing stuff. We could just wrap all the new parsing code in try/catch blocks and fall back to the default values if they fail. That might be a cleaner/simpler approach. 

Please let me know when you've finished adding those tests and making any other changes and I'll give it another review. Please also be sure to run `yarn lint` and `yarn test` (see [`CONTRIBUTING.md`](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md) for details).

bockoblur:
@Hopding If we are using `exifr`, I don't see how we could avoid async, since their parsing method is async. And since every function in which we will be using our new `.resolution()` is already async, I believe it's no big deal to be async. Do you agree?

Hopding:
Both of the methods that load the embedders are async:
https://github.com/Hopding/pdf-lib/blob/f472c0a50fa2f59e5fbc68139ce37558afbb1255/src/core/embedders/JpegEmbedder.ts#L31
https://github.com/Hopding/pdf-lib/blob/f472c0a50fa2f59e5fbc68139ce37558afbb1255/src/core/embedders/PngEmbedder.ts#L11
So we could run the async logic to compute the resolution in those methods (with try-catches to prevent failures). That wouldn't require anything to be lazyily-run or introduce new async APIs.

I don't think there's anything wrong with async APIs per se. They're inevitable and useful for JS code. But I do think it's worth avoiding them if they aren't necessary, because they do complicate things. And in this case, we could bunch all the async logic together instead of requiring the user to call new async methods. Another thing that must be considered is API consistency. The two existing methods (`scale()` and `size()`, along with `width` and `height` properties) on `PDFImage` that relate to image metadata are synchronous, so it'd be less consistent to make `resolution()` an async once. And if we compute the resolution in the load methods, we could even make `resolution` a simple property (which I think would be quite nice).

All that being said, I'm not strictly opposed to the implementation you currently have. Just want to explore our options and build the cleanest/simplest API we can for this. Once it's released it becomes a real hassle to change, so I try to be as thorough as possible before merging/releasing 😄

bockoblur:
@Hopding thanks for clarification.
New commit is pushed, containing:

- `resolution` is now a property, not a method
- the whole parsing is wrapped in `try...catch` block
- unit tests for embedders include resolution checking
- integration tests for Node, Deno & Web (test1 was modified, adding page 5 with images drawn at 'natural' resolution, i.e. considering resolution information from images)
  - new asset added (`assets/images/with_physical_dimensions.png`), name is self-explanatory.

Linted, tested.

**Note:** I am getting unresolved dependencies warning on build:
```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
fs (imported by node_modules\exifr\dist\full.esm.mjs)
```

Hopding:
@bockoblur I did some digging to try to understand what's causing the build error that you noted. Afraid I don't have time to fully trace it down, but it seems to have something to do with the [builds `exifr` provides](https://github.com/MikeKovarik/exifr#installation). It looks like we're (by default) using the Node version of the build which tries to import `fs` (and is probably also larger than we need). I tried using the `mini` version (there's also a `lite` one) and that seemed to resolve the error, but results in build issues in the tests. I don't have time atm to resolve those as well, but it looks like it's an issue with how Jest deals with importing ES6 modules. 

Here's the diff of the changes I made:
```diff
diff --git a/src/types/exifr.d.ts b/src/types/exifr.d.ts
new file mode 100644
index 0000000..2bf07a3
--- /dev/null
+++ b/src/types/exifr.d.ts
@@ -0,0 +1,3 @@
+declare module 'exifr/dist/mini.esm.mjs' {
+  export * from 'exifr/index';
+}
diff --git a/src/utils/jpg-resolution.ts b/src/utils/jpg-resolution.ts
index 0b0ab8f..80aa617 100644
--- a/src/utils/jpg-resolution.ts
+++ b/src/utils/jpg-resolution.ts
@@ -1,4 +1,4 @@
-import { parse } from 'exifr';
+import { parse } from 'exifr/dist/mini.esm.mjs';
 
 // default resolution
 const DEFAULT_RESOLUTION = 72;
```

Also, it looks like you accidentally committed a `package-lock.json` file. Would you please remove this? The `yarn.lock` is the only lock file we want committed to the repo.

bockoblur:
@Hopding It seems that build _warning_ from `rollup `can be safely ignored, because all integration test are passing anyway.
So I changed `rollup.config.js` to [ignore this warning](https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency) for now, and we have a clean build record now.  

As for the bundle size, we can't use `lite` version because it doesn't include what we need, and `mini` is not that much smaller than `full` anyway. 

See if this is ok with you.

Hopding:
@bockoblur Unfortunately we can't ignore this error. It's true that it does not affect Node, Deno, or Browser environments. But it does affect React Native. When I try to build the React Native tester app on this branch, I get the following error in the packager logs:
```
error: bundling failed: Error: Unable to resolve module `fs` from `/Users/user/github/pdf-lib/apps/rn/node_modules/pdf-lib/node_modules/exifr/dist/full.umd.js`: Module `fs` does not exist in the Haste module map
```
React Native can be a hassle when it comes to imports. But we can't ignore it since pdf-lib is designed to support all modern JavaScript environments. I don't think the error can be resolved via React Native build configuration changes. But even if it could, I do not want users to have to hassle with that. It's important that pdf-lib Just Work™ out of the box.

-----

Also, I'm a bit worried about adding a 70kb dependency (increasing pdf-lib's bundle size by 16%) just to support getting JPEG resolutions. Even if we used `ts-exif-parser` (which I have other concerns about unrelated to bundle size), we would be adding a 25kb dependency (increasing pdf-lib's bundle by 5%). 

Considering all this, and the trouble we're having with these dependencies, I have some reservations about adding this feature to pdf-lib. Perhaps it would be better to write up a Gist demonstrating how to do this and link to it in pdf-lib's README? 

Sorry, I hate to backpedal at this point. Somebody recently filed an issue about the size of pdf-lib and unused dependencies (https://github.com/Hopding/pdf-lib/issues/522) which brought pdf-lib's bundle size to the front of my mind again.

bockoblur:
Your arguments are spot on.
I will try to write minimal exif parser for jpeg on my own when I find more time.
Cheers.

Hopding:
@bockoblur I’m glad we’re in agreement! And thanks again for the time you’ve spent on this. I know how frustrating it is to keep running into unexpected issues after sinking a lot of effort into something. 

Looking forward to seeing what you come up with for a minimal EXIF parser 👍

Hopding:
I'm going to close this PR for now. Happy to reopen whenever you get around to implementing those changes.

PR #511, opened 2020-07-02
Rotation for ellipse and svgPath

soadzoor:
This PR adds high-level support for rotating svgPaths, and ellipses.

Some notes for @Hopding :
- I couldn't figure out why the rotating didn't have any effect on the ellipse on my first attempt, but I went on, and implemented the same thing for svgPaths.

- For svgPaths, it worked as expected. So, as a workaround/fix, I've changed the ellipse-drawing method to use the `drawSvgPath` as well: it uses the same bezier curves as before, but they're now defined in the path string. I think it's even better this way, because the codebase got a little bit smaller (no need for the `drawEllipse` methods anymore).

- I've modified some tests: test1, and test12 now includes rotation for ellipse, and for svgPath respectively.

- There's a "break" in functionality though: in test1, when you create that "mask" effect for the text with the 2 ellipses. That no longer works as expected. There seems to be a pretty huge offset for some reason, but I couldn't figure out the solution. Some additional notes about this:

  - If I give proper colors to the ellipses (instead of undefined), they appear on the right position and size
  - If I change these 2 lines in `PDFPage.ts` from this:
  ```
  const x = -xScale;
  const y = -yScale;
  ```
  to this:
  ```
   const x = centerX - xScale;
   const y = centerY - yScale;
   ```
  Then the "mask" effect seems to be fine, but all the visible ellipses are having an offset. I think it must be related to something like a "context position" that is being (or needs to be) modified (with `page.moveTo`), or something. Because the original ellipse drawing methods used something like that. And this bring me to my last note:
  - I tried to create the same "mask" effect with rectangles instead of ellipses, but it didn't work for me neither with, nor without this PR. So I think it was already a quite unique, "hacky" (or at least low-level) solution before, that it only worked with ellipses.

### DIFF
--- a/apps/web/test12.html
+++ b/apps/web/test12.html
@@ -116,7 +116,7 @@
       const page = pdfDoc.addPage(PageSizes.Letter);
 
       // quadratic bezier example
-      page.drawSvgPath('M0,0 q100,125 100,0 T200,0', {
+      page.drawSvgPath('M200,300 Q400,50 600,300 T1000,300', {
         x: 100,
         y: 700,
         scale: 0.5,
--- a/apps/web/test12.html
+++ b/apps/web/test12.html
@@ -116,13 +116,13 @@
       const page = pdfDoc.addPage(PageSizes.Letter);
 
       // quadratic bezier example
-      page.drawSvgPath('M200,300 Q400,50 600,300 T1000,300', {
-        x: inchToPt(-1),
-        y: inchToPt(11),
-        scale: 0.5,
+      page.drawSvgPath('M0,0 q100,125 100,0 T200,0', {
+        x: 100,
+        y: 700,
         borderWidth: 4,
         borderDashArray: [24, 12],
         borderLineCap: LineCapStyle.Round,
+        rotate: degrees(-30),
       });
       page.drawSvgPath('M200,300 L400,50 L600,300 L800,550 L1000,300', {
         x: inchToPt(-1),
--- a/apps/web/test12.html
+++ b/apps/web/test12.html
@@ -119,10 +119,11 @@
       page.drawSvgPath('M0,0 q100,125 100,0 T200,0', {
         x: 100,
         y: 700,
+        scale: 0.5,
         borderWidth: 4,
         borderDashArray: [24, 12],
         borderLineCap: LineCapStyle.Round,
-        rotate: degrees(-30),
+        rotate: degrees(30),
       });
       page.drawSvgPath('M200,300 L400,50 L600,300 L800,550 L1000,300', {
         x: inchToPt(-1),
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -1427,9 +1427,8 @@ export default class PDFPage {
    * @param options The options to be used when drawing the ellipse.
    */
   drawCircle(options: PDFPageDrawCircleOptions = {}): void {
-    let { size } = options;
+    const { size = 100 } = options;
     assertOrUndefined(size, 'size', ['number']);
-    size = asNumber(size ?? 100);
     this.drawEllipse({ ...options, xScale: size, yScale: size });
   }
 
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -52,7 +52,6 @@ import {
   assertRangeOrUndefined,
   assertIsOneOfOrUndefined,
 } from 'src/utils';
-import {asNumber} from "./objects";
 
 /**
  * Represents a single page of a [[PDFDocument]].
@@ -1378,6 +1377,25 @@ export default class PDFPage {
     assertOrUndefined(options.borderColor, 'options.borderColor', [
       [Object, 'Color'],
     ]);
+    assertRangeOrUndefined(
+      options.borderOpacity,
+      'options.borderOpacity',
+      0,
+      1,
+    );
+    assertOrUndefined(options.borderWidth, 'options.borderWidth', ['number']);
+    assertOrUndefined(options.borderDashArray, 'options.borderDashArray', [
+      Array,
+    ]);
+    assertOrUndefined(options.borderDashPhase, 'options.borderDashPhase', [
+      'number',
+    ]);
+    assertIsOneOfOrUndefined(
+      options.borderLineCap,
+      'options.borderLineCap',
+      LineCapStyle,
+    );
+    assertIsOneOfOrUndefined(options.blendMode, 'options.blendMode', BlendMode);
     const graphicsStateKey = this.maybeEmbedGraphicsState({
       opacity: options.opacity,
       borderOpacity: options.borderOpacity,
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -1425,7 +1425,9 @@ export default class PDFPage {
       y: centerY,
       rotate: options.rotate,
       color: options.color,
+      opacity: options.opacity,
       borderColor: options.borderColor,
+      borderOpacity: options.borderOpacity,
       borderWidth: options.borderWidth,
       borderDashArray: options.borderDashArray,
       borderDashPhase: options.borderDashPhase,
@@ -1452,8 +1454,9 @@ export default class PDFPage {
    * @param options The options to be used when drawing the ellipse.
    */
   drawCircle(options: PDFPageDrawCircleOptions = {}): void {
-    const { size } = options;
+    let { size } = options;
     assertOrUndefined(size, 'size', ['number']);
+    size = asNumber(size ?? 100);
     this.drawEllipse({ ...options, xScale: size, yScale: size });
   }
 
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -1,6 +1,5 @@
 import { Color, rgb } from 'src/api/colors';
 import {
-  drawEllipse,
   drawImage,
   drawLine,
   drawLinesOfText,
@@ -52,6 +51,7 @@ import {
   assertRangeOrUndefined,
   assertIsOneOfOrUndefined,
 } from 'src/utils';
+import {asNumber} from "./objects";
 
 /**
  * Represents a single page of a [[PDFDocument]].
@@ -1126,6 +1126,7 @@ export default class PDFPage {
     assertOrUndefined(options.x, 'options.x', ['number']);
     assertOrUndefined(options.y, 'options.y', ['number']);
     assertOrUndefined(options.scale, 'options.scale', ['number']);
+    assertOrUndefined(options.rotate, 'options.rotate', [[Object, 'Rotation']]);
     assertOrUndefined(options.borderWidth, 'options.borderWidth', ['number']);
     assertOrUndefined(options.color, 'options.color', [[Object, 'Color']]);
     assertRangeOrUndefined(options.opacity, 'opacity.opacity', 0, 1);
@@ -1167,6 +1168,7 @@ export default class PDFPage {
         x: options.x ?? this.x,
         y: options.y ?? this.y,
         scale: options.scale,
+        rotate: options.rotate ?? degrees(0),
         color: options.color ?? undefined,
         borderColor: options.borderColor ?? undefined,
         borderWidth: options.borderWidth ?? 0,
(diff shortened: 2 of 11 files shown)


soadzoor:
Hey @Hopding ,

As you requested: the drawing method is now reverted to the "standard" one for ellipses, and I also figured out the rotation problem, it seems. Not exactly sure of what the problem was, I basically followed how the rectangle is being drawn. Maybe it has to do something with the lack of `translate` operator in the original method. Anyway, it works now. Please check and let me know!

src/api/PDFPage.ts
```diff
@@ -1441,8 +1427,9 @@ export default class PDFPage {
    * @param options The options to be used when drawing the ellipse.
    */
   drawCircle(options: PDFPageDrawCircleOptions = {}): void {
-    const { size } = options;
+    let { size } = options;
     assertOrUndefined(size, 'size', ['number']);
+    size = asNumber(size ?? 100);
```
Hopding, line 1432: If we want to set the default value for `size` explicitly there's a slightly cleaner way to do that:
```suggestion
    const { size = 100 } = options;
    assertOrUndefined(size, 'size', ['number']);
```

src/api/operations.ts
```diff
@@ -27,10 +25,12 @@ import {
   setLineCap,
   setGraphicsState,
   setDashPattern,
+  appendBezierCurve,
 } from 'src/api/operators';
 import { Rotation, toRadians } from 'src/api/rotations';
 import { svgPathToOperators } from 'src/api/svgPath';
 import { PDFHexString, PDFName, PDFNumber, PDFOperator } from 'src/core';
+import {asNumber} from "./objects";
```
Hopding, line 33: Can we revert this import to what's in `master`?
```suggestion
import { asNumber } from 'src/api/objects';
```

src/api/operations.ts
```diff
@@ -236,13 +238,13 @@ export const drawEllipsePath = (config: {
   const ym = y + yScale;
 
   return [
-    pushGraphicsState(),
+    translate(centerX, centerY),
```
Hopding, line 241: Can we leave the `pushGraphicsState()` and `popGraphicsState()` operators in place? If not the graphics state can be unintentionally altered even after this function returns.

src/api/PDFPage.ts
```diff
@@ -1369,38 +1372,20 @@ export default class PDFPage {
...
-    assertOrUndefined(options.borderDashPhase, 'options.borderDashPhase', [
-      'number',
-    ]);
-    assertIsOneOfOrUndefined(
-      options.borderLineCap,
-      'options.borderLineCap',
-      LineCapStyle,
-    );
-    assertIsOneOfOrUndefined(options.blendMode, 'options.blendMode', BlendMode);
-
```
Hopding, line 1396: Any reason for removing these assertions?

src/api/operations.ts
```diff
@@ -219,14 +219,16 @@ export const drawEllipsePath = (config: {
   y: number | PDFNumber;
   xScale: number | PDFNumber;
   yScale: number | PDFNumber;
-}): PDFOperator[] => {
-  let x = asNumber(config.x);
-  let y = asNumber(config.y);
+  rotate: Rotation;
```
Hopding, line 222: This must be marked as optional or else it will be a breaking change.
```suggestion
  rotate?: Rotation;
```

src/api/operations.ts
```diff
@@ -251,6 +253,7 @@ export const drawEllipse = (options: {
   y: number | PDFNumber;
   xScale: number | PDFNumber;
   yScale: number | PDFNumber;
+  rotate: Rotation;
```
Hopding, line 256: This must be marked as optional or it will be a breaking change.
```suggestion
  rotate?: Rotation;
```

src/api/operations.ts
```diff
@@ -272,22 +275,25 @@ export const drawEllipse = (options: {
...
     popGraphicsState(),
   ].filter(Boolean) as PDFOperator[];
 
+
 export const drawSvgPath = (
   path: string,
   options: {
     x: number | PDFNumber;
     y: number | PDFNumber;
+    rotate: Rotation;
```
Hopding, line 296: This also must be marked as optional:
```suggestion
    rotate?: Rotation;
```

apps/web/test12.html
```diff
@@ -116,13 +116,14 @@
       const page = pdfDoc.addPage(PageSizes.Letter);
 
       // quadratic bezier example
-      page.drawSvgPath('M200,300 Q400,50 600,300 T1000,300', {
-        x: inchToPt(-1),
-        y: inchToPt(11),
+      page.drawSvgPath('M0,0 q100,125 100,0 T200,0', {
```
Hopding, line 119: Was there a reason we needed to change this SVG path? I believe the original path came from https://svgwg.org/svg2-draft/paths.html, so it would be nice to keep that one unless we really need to change it.

src/api/PDFPage.ts
soadzoor: Great, I didn't even know about this! :)

src/api/PDFPage.ts
soadzoor: I think it was when I switched to the "drawSvgPath" method inside the "drawEllipse", because "drawSvgPath" already had these assertions. But you're right, since we reverted that change, I'd need to revert this as well

apps/web/test12.html
soadzoor: I just wanted to try how the change of the "movement" command (`M200,300` to `M0,0`) affects the origin (pivot) of the rotation. I've reverted this change now, as well. Let me know if there's anything else!

src/api/operations.ts
soadzoor: I thought the `drawEllipsePath` should be called internally only. If a "user" wants to draw an ellipse, they'll use `drawEllipse` anyway, which already has push, and pop for graphicsstate. I've removed the "export" from `drawEllipsePath` to make sure it is only called within the `operations.ts`. If I re-add the push and pop, it breaks functionality: ellipses are drawn to a different location than you'd expect.

Although I'm not 100% sure of the functionalities / effects of graphicsstates, I think it wouldn't make sense to push 2 graphicsstates for drawing one ellipse. We use only one push, and one pop for all the rest of the draw functions (svgPath, rectangle, image, etc). I'm starting to think that maybe this double-push-pop of graphic states was the reason why rotation didn't work for the first time.

Hopding: I'm afraid we can't remove the `export` for this (or any) function since releases have already been cut that include the `export`. Removing the `export` would be a breaking change, and only large refactorings or big feature additions warrant something like that (although little changes can be included in those releases), because they force users to refactor their code.

That's actually another reason why we have to leave the push and popping operators in. Removing them would also constitute a breaking change that could mess up the code for users that might be calling `drawEllipsePath`. Though, aside from that, there's nothing wrong with having multiple pushs/pops in a given graphics operation. It's actually a fairly common thing with the kind of state-based graphics model that PDF defines.

I tried adding them back to see what would happen and did not notice a difference in the output PDF. Are you sure that adding them back messes up the rotation? Perhaps that was an issue in a previous version of the code you are thinking of?

soadzoor: I've just created 2 screenshots. The left one is without the additional "push-pop", the right one is with the additional "push-pop" of graphicsstates inside "drawEllipsePath". That's the only difference in the code. So, as you can see, it basically breaks the functionality of drawing ellipses as we know it.

test1


Additionally: I don't know why anyone would use "drawEllipsePath" by itself, as it's clearly just part of the job when you want to draw an ellipse. I'd bet that all of the users who draw ellipses on PDFs with this lib use "drawEllipse", and not "drawEllipsePath". If (for any reason) someone really uses "drawEllipsePath" at the moment, they could just refactor that to "drawEllipse", and it would do the same thing. Except they'd have the opportunity to add additional (but optional) parameters (like colors, opacity, dashed borders, etc). So, I get that it's a breaking change, but it'd be super easy to solve the issues (even easier if the user reads the changelog, and the "migration guide" - if we're about to have one).

I probably can't see the bigger picture, but I just can't see why this is a problem.

Also, if we take a look at "svgPathToOperators", it doesn't "push-pop" graphicsstates either, but it is an exported function. that calls graphics-related operators (like `appendBezierCurve`, `moveTo`, `lineTo`, etc). Why does "drawEllipsePath" need to have "push-pop", if "svgPathToOperators" doesn't need them? To me, it doesn't really make any sense, but again: I probably can't see the bigger picture.

Hopding: Yep, that certainly looks broken to me! I'm still not able to reproduce what you've captured above though. I made the following changes:
```diff
diff --git a/src/api/operations.ts b/src/api/operations.ts
index 177ef22..d24f55d 100644
--- a/src/api/operations.ts
+++ b/src/api/operations.ts
@@ -220,8 +220,7 @@ const drawEllipsePath = (config: {
   xScale: number | PDFNumber;
   yScale: number | PDFNumber;
   rotate?: Rotation;
-}): PDFOperator[] =>
-{
+}): PDFOperator[] => {
   const centerX = asNumber(config.x);
   const centerY = asNumber(config.y);
   const xScale = asNumber(config.xScale);
@@ -238,6 +237,7 @@ const drawEllipsePath = (config: {
   const ym = y + yScale;
 
   return [
+    pushGraphicsState(),
     translate(centerX, centerY),
     rotateRadians(toRadians(config.rotate ?? degrees(0))),
     moveTo(x, ym),
@@ -245,6 +245,7 @@ const drawEllipsePath = (config: {
     appendBezierCurve(xm + ox, y, xe, ym - oy, xe, ym),
     appendBezierCurve(xe, ym + oy, xm + ox, ye, xm, ye),
     appendBezierCurve(xm - ox, ye, x, ym + oy, x, ym),
+    popGraphicsState(),
   ];
 };
 
@@ -287,7 +288,6 @@ export const drawEllipse = (options: {
     popGraphicsState(),
   ].filter(Boolean) as PDFOperator[];
 
-
 export const drawSvgPath = (
   path: string,
   options: {
```
And when I draw an ellipse with the following:
```js

  const page = pdfDoc.addPage([300, 300]);
  page.drawEllipse({
    x: 100,
    y: 100,
    yScale: 75,
    xScale: 50,
    color: rgb(1, 0, 0),
    rotate: degrees(45),
  });

  const pdfBytes = await pdfDoc.save();
```
the document looks the same whether I add the push/pop operators or not.

But I don't mean to be a stickler and I could certainly just be doing something wrong on my end. What we can do is just create a new function that is not exported and deprecate `drawEllipsePath` (with an `/** @deprecated */` comment). And the new function would just be the version of `drawEllipsePath` you've implemented in this PR.

soadzoor: @Hopding Have you tried test1 (from the integration tests)? That's the one I'm testing with. When I make the same changes you showed me in your last comment (pushGraphicsState, popGraphicsState), it has the effect of what you can see on my screenshot above. Anyway, you're suggestion works for me as well. I've just pushed a PR with it. Thank you for your flexibility!

Hopding:
@soadzoor I was just running through the full suite of integration tests before cutting a new release with this feature. However, while doing so I noticed a problem. Here's what integration test #1 looks like in Adobe Acrobat prior to this PR:

Screen Shot 2020-07-26 at 9 34 19 AM

Of course, the orange ellipse is not rotated here. But note the magenta text overlaid on top of it. Here's what that same test looks like in Acrobat after this PR:

Screen Shot 2020-07-26 at 9 31 27 AM

The ellipse is rotated, but the magenta text is now missing. Interestingly, this does not seems to be the case for all PDF readers. So far I've tested in Acrobat, Foxit, and Preview. The magenta text is missing in Acrobat and Foxit, but not in Preview:

Screen Shot 2020-07-26 at 9 31 11 AM

I usually test PR changes with Preview and only prior to cutting a release do I run the full suite of tests in all readers and environments (because it's slow and time consuming). That's why I didn't catch this until now. 

I haven't had a chance to look into what's causing this. And it's intriguing that not all readers have the same problem. But regardless, I'm postponing the release until you have some time to investigate this. Note that I won't revert the PR merge just yet. But I'll have to do that prior to the next release of other features, unless we're able to merge a fix for this.

soadzoor:
Hey @Hopding ,

I noticed this right away (because I usually test PDFs with chrome's built-in reader). I didn't know that it's still okay in some readers, though.

**See my original post on this PR, from this part: `There's a "break" in functionality though...`, it's all about this issue.**

My guess is that some readers don't respect transformations for the clipping (if that's possible..?). So I believe there's not much we can do about it, because we need transformations for rotating. Note, that we were only able to use ellipses for this overlaying-mask-text-effect (prior to this PR) anyway. It didn't work with any other shapes as far as I know. So it was already a very special, "unstable", "unreliable", and also low-level feature meaning the user needed to do some low-level stuff to get it done. And again: it only worked with ellipses prior this PR.

Now, that ellipses are also created with some transformations (just like the rest of the shapes), it no longer works (at least not with all the pdf readers). It's a pity, but at least it's consistent with the rest of the shapes. And if I got to choose which one is more important:

1. The ability to use low-level code to draw masked texts with ellipses, and with ellipses only.
or
2. The ability to rotate the ellipse with a well-accustomed high-level method (just like the rest of the shapes)

I'd go with the second option, because:
- It's most likely a feature that more users need. For one example, me and my team, we're long waiting for the second ability, and we'll never use the first.
- It also makes the lib more consistent (ellipses would behave much more like the rest of the shapes). No special treatment / additional functionality anymore.

Considering all the above, I wouldn't hesitate much.

Of course, it's your decision, not mine.

Hopding:
@soadzoor Thanks for your thoughtful reply. You may well be correct that the current behavior is anomalous and doesn't make sense to try and preserve. If that is the case, I'm not necessarily opposed to releasing this change. However, before I commit to doing so, I need to better understand what exactly is going on here. The interplay between clipping, transformations, and inconsistent behavior between different readers is frustrating and I haven't had the time to investigate this yet.

I also tend to agree with you that having the ability to rotate ellipses would be more useful than the ability to mask with only ellipses. What do you think about the following approach?
1. If no rotation is passed (`rotation === undefined`), emit the same sequence of operators as the current version (probably using `drawEllipsePath`).
2. If a rotation is passed, use the new `drawEllipseCurves` function that you've introduced in this PR.

Implementing this approach would amount to a simple ternary condition in the `drawEllipse` function. 

This approach would avoid introducing a breaking change, so we would not need to delay the release of this feature any longer. Admittedly, it's a bit awkward and may seem like overkill. But the only alternatives that I see are:
* I release this breaking change without fully understanding what's going on here, which makes me uneasy. 
* I delay the release of this feature until I have to time to investigate what's happening and determine if a breaking change is the right solution, which might take several weeks.

My apologies for the wait you and your team have had to put up with to get what would seem to be an obvious, simple feature. I appreciate your attentiveness and persistence in getting this feature delivered!

soadzoor:
@Hopding Thank you for your respone! I'm absolutely fine with your proposed approach, thank you! It seems like a decent solution for now, and I hope you'll get a chance to get to the bottom of this in the future! It surely seems like a frustratig anomaly

Hopding:
Version [`1.10.0`](https://www.npmjs.com/package/pdf-lib/v/1.10.0) is now published with this feature. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.10.0).

```
npm install pdf-lib@1.10.0
```

It's also available on unpkg:

- https://unpkg.com/pdf-lib@1.10.0/dist/pdf-lib.js
- https://unpkg.com/pdf-lib@1.10.0/dist/pdf-lib.min.js

As well as jsDelivr:

- https://cdn.jsdelivr.net/npm/pdf-lib@1.10.0/dist/pdf-lib.js
- https://cdn.jsdelivr.net/npm/pdf-lib@1.10.0/dist/pdf-lib.min.js

soadzoor:
Thank you @Hopding  ! For documenting purposes, could you perhaps add the "add rotation to `drawSvgPath`" to the release notes (beside the "add rotation to `drawEllipse`")? Thanks

Hopding:
@soadzoor good catch! I forgot to mention that. I'll get it added to the release notes 👍

PR #513, opened 2020-07-02
Feature/new drawing operators

taxilian:
- Add drawLines helper
- Add borderRadius to drawRectangle to allow rounded corner rectangles


src/api/PDFFont.ts
```diff
@@ -7,6 +7,18 @@ import {
...
+export interface FontMetrics {
+  /** The font's ascender */
+  ascent: number | void;
+  /** The font's descender */
+  descent: number | void;
+  /** The font's line gap if known; only works on custom fonts currently */
+  lineGap: number;
+  /** Font bounding box */
+  bbox: BoundingBox,
+}
```
Hopding, line 21: Please use `undefined` (or, better, an optional property `?:`) instead of `void`. Also, `lineGap` should be optional as well, since it is not present on custom fonts. Users may choose to default the value to zero if they wish, but pdf-lib should not hide the fact that it isn't always available from them.
```suggestion
export interface FontMetrics {
  /** The font's ascender */
  ascent?: number;
  /** The font's descender */
  descent?: number;
  /** The font's line gap, if it's accessible (currently only accessible on custom fonts) */
  lineGap?: number;
  /** Font bounding box */
  bbox: BoundingBox,
}
```

src/api/PDFFont.ts
```diff
@@ -89,6 +101,13 @@ export default class PDFFont implements Embeddable {
     return this.embedder.widthOfTextAtSize(text, size);
   }
 
+  /**
+   * Returns basic metrics about the font
+   */
+  getFontMetrics() : FontMetrics {
+    return this.embedder.getFontMetrics();
+  }
```
Hopding, line 109: Since these font metrics are immutable, I'd prefer to set them as properties of the `PDFFont` (like what we do with `fontName`) rather than add a new method to access them. These fields can then be set in the constructor.

src/core/embedders/CustomFontEmbedder.ts
```diff
@@ -82,6 +82,15 @@ class CustomFontEmbedder {
...
 
+  getFontMetrics() {
+    return {
+      ascent: this.font.ascent,
+      descent: this.font.descent,
+      lineGap: this.font.lineGap,
+      bbox: this.font.bbox,
+    };
+  }
+
```
Hopding, line 93: Please see my previous comment about setting fields in the constructor instead of adding a new method. I think the same thing applies to the font embedders themselves.

src/core/embedders/StandardFontEmbedder.ts
```diff
@@ -81,6 +82,21 @@ class StandardFontEmbedder {
     return (1000 * height) / (yTop - yBottom);
   }
 
+  getFontMetrics() {
```
Hopding, line 85: Please see my previous comment.

src/core/embedders/StandardFontEmbedder.ts
```diff
@@ -81,6 +82,21 @@ class StandardFontEmbedder {
...
+    const bbox = this.font.FontBBox;
+    return {
+      ascent: this.font.Ascender,
+      descent: this.font.Descender,
+      // It would be nice to figure out how to get this value
+      lineGap: 0,
```
Hopding, line 91: We should set this value to `undefined` rather than `0` so users can tell when it is actually `0` vs simply not being available.

src/api/PDFFont.ts
```diff
@@ -7,6 +7,18 @@ import {
...
+
+export interface FontMetrics {
+  /** The font's ascender */
+  ascent: number | void;
+  /** The font's descender */
+  descent: number | void;
+  /** The font's line gap if known; only works on custom fonts currently */
+  lineGap: number;
+  /** Font bounding box */
+  bbox: BoundingBox,
```
Hopding, line 20: Instead of using the `BoundingBox` model from `fontkit`, I think it would be best to use the "rectangle" model that is used throughout pdf-lib. This keeps things consistent. That model is as follows:
```ts
{ x: number, y: number, width: number, height: number }
```

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
```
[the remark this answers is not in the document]
taxilian: editorconfig doesn't take the place of prettier, it gives a cross-IDE way to configure things like tab configuration. Without that I have to reconfigure my editor not to have it indent things incorrectly. It's a fairly widely used tool -- most editors support a plugin for it.

src/api/operations.ts
```diff
@@ -181,36 +183,90 @@ export const drawRectangle = (options: {
...
+    translate(x, y),
     rotateRadians(toRadians(options.rotate)),
     skewRadians(toRadians(options.xSkew), toRadians(options.ySkew)),
-    moveTo(0, 0),
-    lineTo(0, options.height),
-    lineTo(options.width, options.height),
-    lineTo(options.width, 0),
+    moveTo(r, 0),
+    lineTo(w - r, 0),
+    r > 0 ? appendBezierCurve(w - c, 0, w, c, w, r) : null,
```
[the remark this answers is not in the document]
taxilian: I have no strong preference on null vs undefined. Regarding the conditionals, I prefer not to save trivial code paths at the expense of making the output PDF larger by putting completely unneeded operations on it. Granted it's a very small cost, but to my mind it's a larger cost than adding a few branching operations in the code -- particularly given that the cost of executing the bezier function is almost certainly greater than the cost of executing those conditionals.

src/api/operations.ts
```diff
@@ -181,36 +183,90 @@ export const drawRectangle = (options: {
...
+    translate(options.points[0].x, options.points[0].y),
+    rotateRadians(toRadians(options.rotate)),
+    skewRadians(toRadians(options.xSkew), toRadians(options.ySkew)),
+    ...(options.points.slice(1).map(p =>
+          lineTo(
+            asNumber(p.x) - asNumber(options.points[0].x),
+            asNumber(p.y) - asNumber(options.points[0].y),
+    ))),
+    // Close only if instructed to
+    ...(options.closePath ? [closePath()] : []),
```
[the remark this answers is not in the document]
taxilian: It's your project, but I would disagree. Most cases where I would use this function I would neither want nor expect that it would close the path automatically; the only reason I even added that option is so that it would also be a viable replacement for "drawPolygon" in other libraries. The function name is "drawLines" because drawing a number of connected lines is a common operation; if you want it to close path by default then it should be called "drawPolygon", but having it able to not closePath by default seems counter-intuitive.

Your call, but I personally would find the idea that "drawLines" would connect the final point and the first point without being explicitly told to do so very strange.

src/api/operations.ts
```diff
@@ -181,36 +183,90 @@ export const drawRectangle = (options: {
...
-  : closePath(),
+    : options.color                      ? fill()
+    : options.borderColor                ? stroke()
+    : null,
 
     popGraphicsState(),
   ].filter(Boolean) as PDFOperator[];
+};
+
+export const drawLines = (options: {
```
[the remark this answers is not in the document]
taxilian: See my other comment for why I don't think that should be done.

Perhaps a counter-suggestion: add a drawPolygon which calls drawLines and defaults closePaths to true?

I added this function because I'm using it for several different things, including drawing checkmarks and partial outlines -- none of the cases I am using would "drawPolygon" be an appropriate name for what I am doing and in none of them would closing the path be desired.

src/api/PDFFont.ts
taxilian: I'm a bit confused; the custom fonts is where I *have* seen lineGap present, standard fonts does not seem to have them. If not specified the line gap would be 0, since there is then not a gap, so I feel that is the appropriate way to handle it.

src/api/PDFFont.ts
taxilian: I am happy to do that, but I would point out that if you set them as properties then you are executing this far more than if you only use it when you call the method -- there aren't many cases where you need those metrics, so by adding it as a method there is almost no cost to adding it, but if you add it to the constructor and set it as a property then we're setting it every single time the class is instantiated and using the extra memory.

Granted, it's a fairly small cost, but as you seem to be concerned with optimization I just thought I'd point out that putting it in a property is actually less optimal than how I did it =]

src/core/embedders/StandardFontEmbedder.ts
taxilian: My reason for setting it to 0 is that a line gap of 0 will tend to be reasonable for nearly all fonts -- in fact, it's very likely that it's reasonable for all of the standard fonts. It is in fact the line gap which pdf-lib is using as far as I can tell, since I don't believe you are adding any arbitrary line gap.

src/api/PDFFont.ts
taxilian: Can do.

taxilian: Can I ask why you define `{ x: number; y: number; width: number; height: number } ` all over the code instead of actually making it a named interface that can be reused? it would make using the library much nicer...

.editorconfig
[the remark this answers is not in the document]
Hopding: I have no problem with editorconfig per se. I've used it before for projects in other languages. But I don't see what value it adds on top of prettier. Prettier already controls all of the settings specified in this `.editorconfig` (with different values for some of the config). Many editors provide [prettier plugins](https://prettier.io/docs/en/editors.html) that should solve the indentation problems you ran into. It seems preferable to have contributors configure their IDEs to use prettier plugins rather than add additional config.

src/api/PDFFont.ts
Hopding: Appreciate the observations about memory and performance. Performance is indeed something I am concerned about. But there are a number of priorities that have to be balanced. And in this case I think consistency trumps any potential performance issues.

src/core/embedders/StandardFontEmbedder.ts
Hopding: That seems reasonable to me 👍

src/api/PDFFont.ts
Hopding: No particular reason. I wouldn't mind having a `Rectangle` interface.

.editorconfig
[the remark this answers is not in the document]
taxilian: Finally getting back to this a bit; AFAIK prettier only applies things after the fact, whereas editorconfig tells the editor which settings to use for indentation while editing, no? I have prettier plugins installed for vscode and it doesn't seem to configure things in a useful way, but perhaps I'm missing something

Hopding:
@taxilian apologies for my late response. I've been very busy these past few months taking care of some personal stuff and starting my new job at GitHub 🙂 

Is this PR still valid? If so, you'll need to either update this PR (via a force push) or open a new PR with your changes. The tl;dr is that I recently had to do a fairly extensive git history rewrite that caused a bunch of conflicts. Please read https://github.com/Hopding/pdf-lib#git-history-rewrite for details. Apologies for the hassle!

Issue #537, opened 2020-07-23
Different binary for same pdf written multiple times makes regression testing more intensive


trevtrich:
When attempting to regression test pdfs it's nice to be able to look at the binary data to see if anything has changed before attempting to do any more detailed diffing with other, slower, tools. We recently started using pdf-lib to modify our pdfs (excellent work, BTW. it's a great tool) and realized that even when reading and writing the same pdf multiple times, the binaries differ between the two. I did some digging into the content and didn't see any clear path to do something like "ignore header bytes" or something similar. Is there any chance an easy update could be done so that the same pdf written multiple times produces the same binary?

Hopding:
Hello @trevtrich!

By default, when a `PDFDocument` is loaded (or created) its modification date is updated to the current time:

https://github.com/Hopding/pdf-lib/blob/a118cd4cfa1e54622c605385aec34095e79095e9/src/api/PDFDocument.ts#L205

https://github.com/Hopding/pdf-lib/blob/a118cd4cfa1e54622c605385aec34095e79095e9/src/api/PDFDocument.ts#L1160-L1171

This is probably the reason for your binary output being different. However, you can tell `pdf-lib` not to update this metadata by setting the [`updateMetadata`](https://pdf-lib.js.org/docs/api/interfaces/loadoptions#optional-updatemetadata) option to `false`:

```js
const pdfDoc = await PDFDocument.load(..., { updateMetadata: false });
```

This will make the result of `pdfDoc.save()` entirely deterministic based on the input PDF provided and any operations you might have performed after loading it. 

Here's an example script showing how the output of `pdfDoc.save()` is different if you wait one second between loading two documents.

```js
import { PDFDocument } from 'pdf-lib';

const arraysAreEqual = (a: Uint8Array, b: Uint8Array) => {
  if (a.length !== b.length) return false;
  for (let idx = 0, len = a.length; idx < len; idx++) {
    if (a[idx] !== b[idx]) return false;
  }
  return true;
};

const sleep = (millis: number) =>
  new Promise((resolve) => setTimeout(resolve, millis));

(async () => {
  const inputBytes = await fetch(
    'https://pdf-lib.js.org/assets/normal.pdf',
  ).then((res) => res.arrayBuffer());

  const pdfDoc1 = await PDFDocument.load(inputBytes);
  const pdfBytes1 = await pdfDoc1.save();

  await sleep(1000);

  const pdfDoc2 = await PDFDocument.load(inputBytes);
[...]

PR #543, opened 2020-07-28
Allow custom font name on embed

ztoben:
As discussed in #537 this allows the use of a custom font name so that we can have a predictive binary output when embedding fonts into the pdf. Happy to make any changes you deem necessary.

### DIFF
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -792,20 +792,20 @@ export default class PDFDocument {
     font: StandardFonts | string | Uint8Array | ArrayBuffer,
     options: EmbedFontOptions = {},
   ): Promise<PDFFont> {
-    const { subset = false, customFontName = '' } = options;
+    const { subset = false, customName } = options;
 
     assertIs(font, 'font', ['string', Uint8Array, ArrayBuffer]);
     assertIs(subset, 'subset', ['boolean']);
 
     let embedder: CustomFontEmbedder | StandardFontEmbedder;
     if (isStandardFont(font)) {
-      embedder = StandardFontEmbedder.for(font);
+      embedder = StandardFontEmbedder.for(font, customName);
     } else if (canBeConvertedToUint8Array(font)) {
       const bytes = toUint8Array(font);
       const fontkit = this.assertFontkit();
       embedder = subset
-        ? await CustomFontSubsetEmbedder.for(fontkit, bytes, customFontName)
-        : await CustomFontEmbedder.for(fontkit, bytes, customFontName);
+        ? await CustomFontSubsetEmbedder.for(fontkit, bytes, customName)
+        : await CustomFontEmbedder.for(fontkit, bytes, customName);
     } else {
       throw new TypeError(
         '`font` must be one of `StandardFonts | string | Uint8Array | ArrayBuffer`',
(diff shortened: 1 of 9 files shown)


src/api/PDFDocument.ts
```diff
@@ -792,7 +792,7 @@ export default class PDFDocument {
     font: StandardFonts | string | Uint8Array | ArrayBuffer,
     options: EmbedFontOptions = {},
   ): Promise<PDFFont> {
-    const { subset = false } = options;
+    const { subset = false, customFontName = '' } = options;
```
Hopding, line 795: 1. Can we remove the default value for this options? I'd prefer to pass it down as `undefined` since the embedders support that type.
2. Can we rename this to `customName`?

tests/core/embedders/CustomFontEmbedder.spec.ts
```diff
@@ -25,6 +25,16 @@ describe(`CustomFontEmbedder`, () => {
     expect(embedder.fontName).toBe('Ubuntu');
   });
 
+  it(`can set a custom font name`, async () => {
```
Hopding, line 28: In addition to this test, I think it would be good to create a simple [`PDFDocument` test](https://github.com/Hopding/pdf-lib/blob/master/tests/api/PDFDocument.spec.ts) that embeds a custom font with a `customName` and verifies that it serializes to the same value every time it's saved.

ztoben:
@Hopding I think I addressed all of your comments. Let me know if there's any other changes to be made.

src/core/embedders/StandardFontEmbedder.ts
```diff
@@ -21,21 +21,22 @@ export interface Glyph {
...
+  private constructor(fontName: FontNames, customName?: string) {
     // prettier-ignore
     this.encoding = (
         fontName === FontNames.ZapfDingbats ? Encodings.ZapfDingbats
       : fontName === FontNames.Symbol       ? Encodings.Symbol
       : Encodings.WinAnsi
     );
     this.font = Font.load(fontName);
-    this.fontName = this.font.FontName;
+    this.fontName = customName || this.font.FontName;
```
Hopding, line 39: In the `CustomFontEmbedder` I noticed we left the `this.fontName = ...` assignment alone and added a new `customName` class member. And then we determine which one to use when we're computing the value for `BaseFont` with a simple `||`. Can we follow the same approach here for the sake of consistency?

tests/api/PDFDocument.spec.ts
```diff
@@ -123,6 +124,43 @@ describe(`PDFDocument`, () => {
     });
   });
 
+  describe(`embedFont() method`, () => {
```
Hopding, line 127: Can we set `updateMetadata` to `false` in both of these tests? I'm worried about the potential for flakiness if we don't.

src/core/embedders/StandardFontEmbedder.ts
ztoben: @Hopding that seems reasonable. I did notice that in the standard font embedder that `embedIntoContext` references the font directly as `this.font.fontName` rather than the `fontName` class member we set above so I'm not sure it was working correctly anyway.

ztoben: @Hopding that seems reasonable. As it turns out the `embedIntoContext` function in the standard font embedder was actually just referencing the font name directly as `this.font.fontName`, so my change would not have worked as it was anyway.

ztoben:
@Hopding okay, let's try this again 😄

Hopding:
This feature will be available in the next release. I'm planning to cut it this weekend.

Hopding:
Version [`1.10.0`](https://www.npmjs.com/package/pdf-lib/v/1.10.0) is now published with this feature. The full release notes are available [here](https://github.com/Hopding/pdf-lib/releases/tag/v1.10.0).

```
npm install pdf-lib@1.10.0
```

It's also available on unpkg:

- https://unpkg.com/pdf-lib@1.10.0/dist/pdf-lib.js
- https://unpkg.com/pdf-lib@1.10.0/dist/pdf-lib.min.js

As well as jsDelivr:

- https://cdn.jsdelivr.net/npm/pdf-lib@1.10.0/dist/pdf-lib.js
- https://cdn.jsdelivr.net/npm/pdf-lib@1.10.0/dist/pdf-lib.min.js

PR #574, opened 2020-08-25
Add deterministic per-document RNG

### LINKED ISSUE #537: Different binary for same pdf written multiple times makes regression testing more intensive

olihensel:
Fixes #537 
Uses seedrandom to create a seeded random number generator for every document instance. This leads to identical binary documents, if the same actions are executed in the same order.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@pdf-lib/standard-fonts": "^0.0.4",
     "@pdf-lib/upng": "^1.0.1",
     "pako": "^1.0.11",
+    "seedrandom": "^3.0.5",
     "tslib": "^1.11.1"
   },
   "devDependencies": {
@@ -94,6 +95,7 @@
     "@types/jest": "^26.0.0",
     "@types/node-fetch": "^2.5.7",
     "@types/pako": "^1.0.1",
+    "@types/seedrandom": "^2.4.28",
     "@zerollup/ts-transform-paths": "^1.7.18",
     "downlevel-dts": "^0.5.0",
     "flamebearer": "^1.1.3",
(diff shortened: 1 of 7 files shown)


Hopding:
Hello @olihensel! Thanks for working on this, and my apologies for taking so long to review your work.

Earlier this year we released a feature allowing users to specify custom font names (see https://github.com/Hopding/pdf-lib/pull/543) and thereby avoid using random number generation. This partially solves the problem for embedded fonts. But it doesn't solve the broader issue, so I like the approach you've taken in this PR.

Before merging this, I'd like to change two things:
* The RNG is currently being stored on the `PDFDocument` object. A better spot to store it is the [`PDFContext`](https://github.com/Hopding/pdf-lib/blob/master/src/core/PDFContext.ts#L46). The context is the center around which all the other objects in pdf-lib (including `PDFDocument`) are built. Storing it here will allow it to be easily accessed from just about anywhere without having to wire it through constructors and function/method calls. 
* I strongly prefer to avoid adding new dependencies to pdf-lib for a number of reasons. So rather than using `seedrandom`, I'd like to add a RNG function to pdf-lib directly (somewhere in `src/utils` - maybe `src/utils/rng.ts`). I'm no expert on RNGs, but we don't need to worry about cryptographic security for our use case. I found a couple short implementations on StackOverflow that could be copy/pasted (note that we'd need to add a few simple unit tests):
  * https://stackoverflow.com/a/424445
  * https://stackoverflow.com/a/47593316

  We could also pull a RNG from the `seedrandom` source itself: https://github.com/davidbau/seedrandom.

Again, apologies for my very delayed response. I look forward to hearing your thoughts!

Hopding:
Closing this for now. I'd still like to add deterministic random number generation to pdf-lib, but don't personally have time to update this PR to address my previous comments.

PR #621, opened 2020-10-06
Update references to Pika - must reference new name "Skypack"

zackdotcomputer:
I was working on getting a proof-of-concept deno app using this library up and running and ran into the issue that pika.dev is now skypack.dev and will not allow new dependencies that point to their old domain name. You can read more at this [deprecation notice](https://www.pika.dev/cdn?skypack), which you now get instead of the code you requested when going to a cdn.pika.dev URL.

While I had the library open, I figured I would help future developers who would run into the same issue if they tried to cargo-cult your sample code. This PR swaps the references in the README and the deno tests to now point at skypack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,7 +846,7 @@ import {
   PDFDocument,
   StandardFonts,
   rgb,
-} from 'https://cdn.skypack.dev/pdf-lib@^1.6.0';
+} from 'https://cdn.skypack.dev/pdf-lib@^1.11.1';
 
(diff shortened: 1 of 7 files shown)


Hopding (approved):
@zackdotcomputer Thanks for updating this! I've no doubt Deno users will find this to be **very** helpful. I'll update the imports in my Medium article as well.

Hopding:
Hmm, the MR check is hanging... Might have to reopen this to get it unstuck

Issue #490, opened 2020-06-15
Unwanted ligatures


zimt28:
I want to use a font from `fonts.google.com` and am running into an issue.

I've included the font (Nunito) like so:
```js
const nunitoBoldUrl =
  "https://fonts.gstatic.com/s/nunito/v12/XRXW3I6Li01BKofA6sKUYevIWzgPDA.woff2";
const nunitoBoldFontBytes = await fetch(nunitoBoldUrl).then((res) => res.arrayBuffer());

const nunitoBold = await pdfDoc.embedFont(nunitoBoldFontBytes);
```

This works, with one exception (there might be more, but this is what I could find):

Printing `fi` adds unwanted spacing after it which shouldn't be there, the Google preview also doesn't show it.

`fi ir fir`:
<img width="352" alt="Screenshot 2020-06-15 at 21 05 12" src="https://user-images.githubusercontent.com/1764689/84696215-6aa82080-af4c-11ea-951b-21dd5e72dcfb.png">

I _think_ the font has a special ligature for `fi`, if I install the font on my system I get the same connected `f` and `i` letters, but without the spacing. Is this an issue with `pdf-kit` or `fontkit`?

fobdy:
The same problem for me with "Source Sans Pro" font

fobdy:
We could try to turn off the ligatures if there was access to the low-level fontkit api
```js
font.layout(lineTxt, { liga:false })
```

Related https://github.com/foliojs/fontkit/issues/170

zimt28:
@fobdy Which file format are you trying to use?

fobdy:
> @fobdy Which file format are you trying to use?

I'm using `.ttf`

Hopding:
Hello @zimt28 @fobdy! As far as I'm aware this is an issue with `fontkit`, though I haven't been able to confirm this. If anybody is interested, it is possible to disable ligatures when performing text layout with `fontkit` (as @fobdy mentioned). You would just need to update the API to pass options here:
https://github.com/Hopding/pdf-lib/blob/193e48ff2ce07d72f8191087adcdd01aa3c48a62/src/core/embedders/CustomFontEmbedder.ts#L52-L57

I'd be happy to accept a PR if anybody is interested in working on this.

Hopding:
[...]

PR #642, opened 2020-10-21
Font features

hefler:
I hope this helps mitigating issue #490.


src/api/PDFDocumentOptions.ts
```diff
@@ -35,4 +35,5 @@ export interface CreateOptions {
 export interface EmbedFontOptions {
   subset?: boolean;
   customName?: string;
+  features?: any[];
```
Hopding, line 38: I prefer to avoid using the `any` type in the pdf-lib codebase. It prevents users from being able to easily understand what types of features can actually be controlled. Can we use a specific type here that defines what features are supported? E.g. `features: { foo: boolean, bar: string }`.

Hopding (changes requested):

src/api/PDFDocumentOptions.ts
hefler: First, let me get it out my chest: I'm new to TS. 🙈

That being said, I've tried to set the type like so: `string[] | {[key: string]: boolean}`, which theoretically is the type that param can be, but  I got errors as fontKit set that type to `any[]` (hence why I used it in the first place). So we could either leave like fontKit set it or choose one type only like `string[]` which I didn't get any errors. What would you suggest?

PS: I can see why they did like that, there's a long list of possible features (https://docs.microsoft.com/en-gb/typography/opentype/spec/featurelist)

Hopding: No worries! Appreciate you contributing despite being new :)

`fontkit` doesn't actually have any TS types. `pdf-lib` uses a fork called `@pdf-lib/fontkit`, which has its own types defined [here](https://github.com/Hopding/fontkit/blob/master/fontkit.d.ts). I created them based on the `fontkit` README, which does not document the ability to pass objects in addition to arrays. We probably need to update the [`@pdf-lib/fontkit/fontkit.d.ts`](https://github.com/Hopding/fontkit/blob/master/fontkit.d.ts) file to accept both objects and arrays (and really they shouldn't be `any` types either, if we know it's just strings and booleans that are allowed). Once that's done, we should be able to use the `string[] | {[key: string]: boolean}` type without any issues.

I would prefer to have an interface that essentially represents the table in https://docs.microsoft.com/en-gb/typography/opentype/spec/featurelist. It would be nice for autocomplete, typechecking, and documentation purposes. However, if you prefer to just have `string[] | {[key: string]: boolean}`, then I'm fine with that too.

package.json
```diff
@@ -87,15 +87,15 @@
...
+    "@pdf-lib/fontkit": "^1.1.0",
     "@rollup/plugin-commonjs": "^13.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^8.0.1",
     "@types/jest": "^26.0.0",
     "@types/node-fetch": "^2.5.7",
     "@types/pako": "^1.0.1",
     "@zerollup/ts-transform-paths": "^1.7.18",
-    "downlevel-dts": "^0.5.0",
+    "downlevel-dts": "^0.6.0",
```
Hopding, line 98: Does this need to be updated as part of this PR?

Hopding (changes requested):
@hefler Can we update one of the integration tests (for all 4 environments) to pass a font feature to disable ligatures (per https://github.com/Hopding/pdf-lib/issues/490)? That way we know this works properly and it won't be broken in the future.

Also, notice that the pipeline for this PR failed. The pipeline must pass before this can be merged. Looks like it was the [`yarn typecheck` command](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-type-checker) that failed.

package.json
hefler: To be honest I don't know, I believe this was caused by an `audit fix`.
I'm also not super familiar writing tests either, but I can give it a go.

Hopding: In that case I'd like to undo this change. If the dependency needs to be updated, I'd prefer that to be a separate PR. But I don't think it's necessary for the changes you're making anyways.

Hopding: Also, with respect to testing, no worries if you're new to it! I'd recommend you read over the [`CONTRIBUTING.md`](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md) file. All contributors should read it. There are sections explaining how to run the [unit tests](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-unit-tests), [integration tests](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-integration-tests) (though you don't need to worry about running the React Native tests), and [running the typechecker](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md#running-the-type-checker) (amongst other important things). Let me know if you have any questions.

Hopding:
@hefler Thanks for working on this! These changes will go out in the next release of pdf-lib!

hefler:
@Hopding No worries! Also thanks for the patience and guidance ✌️

Issue #477, opened 2020-06-09
Embed JS


btecu:
Is there a way to embed Javascript into a PDF using `pdf-lib`?

Hopding:
Hello @btecu! 

You certainly can. However, pdf-lib doesn't have any special APIs for this, so you'll have to use the lower-level APIs. Here's a basic example demonstrating how to create a JavaScript action dictionary with pdf-lib:
```js
import { PDFDocument, PDFHexString } from 'pdf-lib';

const pdfDoc = await PDFDocument.create()

const jsActionDict = pdfDoc.context.obj({
  Type: 'Action',
  S: 'JavaScript',
  JS: PDFHexString.fromText(`
    // Insert the JavaScript you want to embed here...
  `),
})

// Now use the `jsActionDict`...
```

See section **12.6.4.16 JavaScript Actions** of the [PDF spec](https://github.com/Hopding/pdf-lib/blob/master/pdf_specification.pdf) for more details about JavaScript action dictionaries and how to use them.

PR #643, opened 2020-10-22
Add high-level API for adding document JavaScripts

brodo:
This PR adds an API for adding JavaScript to a document. It is based on [this issue](https://github.com/Hopding/pdf-lib/issues/477)).

### DIFF
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -716,11 +716,15 @@ export default class PDFDocument {
   /**
    * Add document JavaScript. The script is executed when the document is opened.
    * See the [JavaScript™ for Acrobat® API Reference](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf)
-   * for details.
-   * @param script The JavaScript to execute.
+   * for details. For example:
+   *
+   * ```js
+   * pdfDoc.addJavascript('main', 'console.show(); console.println("Hello World")');
+   * ```
    * @param name The name of the script. Must be unique per document.
+   * @param script The JavaScript to execute.
    */
-  addJavascript(script: string, name: string) {
+  addJavascript(name: string, script: string) {
     const jsActionDict = this.context.obj({
       Type: 'Action',
       S: 'JavaScript',
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -714,6 +714,29 @@ export default class PDFDocument {
   }
 
   /**
+   * Add document JavaScript. The script is executed when the document is opened.
+   * See the [JavaScript™ for Acrobat® API Reference](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf)
+   * for details.
+   * @param script The JavaScript to execute.
+   * @param name The name of the script. Must be unique per document.
+   */
+  addJavascript(script: string, name: string) {
+    const jsActionDict = this.context.obj({
+      Type: 'Action',
+      S: 'JavaScript',
+      JS: PDFHexString.fromText(script),
+    });
+    const jsActionRef = this.context.register(jsActionDict);
+    const jsNameTree = this.context.obj({
+      Names: this.context.obj([name, jsActionRef]),
+    });
+    this.catalog.set(
+      PDFName.of('Names'),
+      this.context.obj({ JavaScript: jsNameTree }),
+    );
+  }
+
+  /**
    * Add an attachment to this document. Attachments are visible in the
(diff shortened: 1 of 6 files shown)


src/api/PDFDocument.ts
```diff
@@ -713,6 +713,29 @@ export default class PDFDocument {
...
+    const jsActionDict = this.context.obj({
+      Type: 'Action',
+      S: 'JavaScript',
+      JS: PDFHexString.fromText(script),
+    });
+    const jsActionRef = this.context.register(jsActionDict);
+    const jsNameTree = this.context.obj({
+      Names: this.context.obj([name, jsActionRef]),
+    });
+    this.catalog.set(
```
Hopding, line 733: It looks like this will override any existing JavaScript that may exist in the document. We need to add to the existing tree if one exists, otherwise we should create a new one.

src/api/PDFDocument.ts
```diff
@@ -713,6 +713,29 @@ export default class PDFDocument {
...
   }
 
+  /**
+   * Add document JavaScript. The script is executed when the document is opened.
+   * See the [JavaScript™ for Acrobat® API Reference](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf)
+   * for details.
+   * @param script The JavaScript to execute.
+   * @param name The name of the script. Must be unique per document.
+   */
+  addJavascript(script: string, name: string) {
```
Hopding, line 723: Can we alter the implementation of this method a bit? There are some existing patterns in the codebase that should be followed. Please see https://github.com/Hopding/pdf-lib/blob/master/src/api/PDFEmbeddedFile.ts#L48. File embedding is similar to JavaScript embedding in some respects.

src/api/PDFDocument.ts
```diff
@@ -713,6 +713,29 @@ export default class PDFDocument {
     return copiedPages;
   }
 
+  /**
+   * Add document JavaScript. The script is executed when the document is opened.
+   * See the [JavaScript™ for Acrobat® API Reference](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf)
+   * for details.
+   * @param script The JavaScript to execute.
+   * @param name The name of the script. Must be unique per document.
+   */
```
Hopding, line 722: The standard for the pdf-lib codebase is to add example code to all doc comments for each public method. Please add one here. You can reference other methods to see what this should look like.

Hopding (changes requested):
Hello @brodo! 

Thank you very much for working on this. I'm sure many users will find this to be a useful addition to the library. I've just requested a few changes.

Please also update some of the integration and unit tests to exercise this new functionality (I'd recommend taking a look at the [CONTRIBUTING.md](https://github.com/Hopding/pdf-lib/blob/master/CONTRIBUTING.md) if you haven't had a chance already).

src/api/PDFDocument.ts
brodo: Hi! Thanks for the quick feedback and warm welcome! 😊 This should be fixed by the new commits.

src/api/PDFDocument.ts
brodo: I've taken the code you suggested and adopted it. It's way better this way...

src/api/PDFDocument.ts
brodo: I added a short example.

brodo:
I also added unit tests and integration tests for node. As pretty much no PDF viewer besides Acrobat supports this, I think the other integration tests do not make much sense.

src/api/PDFDocument.ts
Hopding: Thanks! I'll try to review it again tomorrow evening.

Hopding:
@brodo Thanks for adding the test! It's true that Acrobat is pretty much the only reader that supports JavaScript in PDFs. However, the tests for Deno, web, and React Native will still need to be updated to exercise this feature. This is because (a) The tests are meant to be exact duplicates between environments (to the greatest extent possible), this makes them much easier to update and maintain. And (b) we need to ensure that the implementation of `addJavascript` works in all environments, regardless of which PDF readers are used to open the resulting PDF. Unfortunately, it's easy to mistakenly implement a feature that only works in Node, or only the browser, etc...

brodo:
Should be done now!

Hopding (approved):
Thanks @brodo! I'll merge this into `master` shortly and it will go out in the next release of `pdf-lib`.

Hopding:
This feature was just released in version `1.12.0`: https://pdf-lib.js.org/docs/api/classes/pdfdocument#addjavascript

PR #677, opened 2020-11-18
Support hexadecimal Default Appearance

btecu:
`getDefaultAppearance` is used to determine font size and the style among other things.
Some pdf files can have the `DA` property as a hexadecimal, causing the library to have a un-parsable `DA` value.

Bonus: Make return value match the function signature (`string | undefined`).

### DIFF
--- a/src/core/acroform/PDFAcroField.ts
+++ b/src/core/acroform/PDFAcroField.ts
@@ -87,7 +87,12 @@ class PDFAcroField {
   }
 
   getDefaultAppearance(): string | undefined {
-    return this.DA()?.asString() ?? '';
+    const DA = this.DA();
+    if (DA instanceof PDFHexString) {
+      return DA.decodeText();
+    }
+
+    return DA?.asString();
   }
 
   getFlags(): number {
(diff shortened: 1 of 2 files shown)


Hopding:
@btecu Do you have an example of a PDF that contains hex string DAs? Also, I'm all for making this change, but I would like to create some unit tests around this method to ensure the functionality isn't broken or removed down the line.

btecu:
@Hopding I sent you an example document.

Hopding:
Thanks @btecu! Would you please create a unit test for each of the altered methods? For example, `tests/core/acroform/PDFAcroField.spec.ts` would load the document you shared with me (thus it would need to be added to the `assets/pdfs` directory), read the raw appearance strings and assert that they are hex strings, and then call the `getDefaultAppearance()` methods and assert that the correct value is returned. Note that we'll need to do this for both `PDFAcroField` and `PDFWidgetAnnotation` instances.

btecu:
@Hopding I cannot add the PDF that I shared with you since is not a public document, and I don't know how to add a test for it without the PDF.

Hopding:
@btecu I added tests in https://github.com/Hopding/pdf-lib/pull/693. These changes will go out in the next release of pdf-lib. Thanks for your contribution! 🎉

Hopding:
The changes in this PR were just released in `v1.13.0`.

PR #687, opened 2020-12-01
Handle negative bounds

### LINKED ISSUE #680: Unhandled bounds
While testing filling forms I noticed a few cases where the text had really small font size.
It turned out that the issue was a negative bound, which apparently is valid.
So likely we need `Math.abs(bounds.height)` here:
https://github.com/Hopding/pdf-lib/blob/581d87e2a12628df94161df72c9d36e661eb7e2c/src/api/text/layout.ts#L50

btecu:
Fixes #680.


Hopding (approved):
Thanks @btecu!

Hopding:
The changes in this PR were just released in `v1.13.0`.

PR #695, opened 2020-12-09
Widget Page Reference

btecu:
Exposing the page reference allows for finding the page a widget is on:
```js
for (let widget of pdfField.getWidgets()) {
  let page = document.getPages().find(x => x.ref === widget.P());
  // Do work with the page
}
```

Example of PDF file that has widgets with page references:
https://www.irs.gov/pub/irs-pdf/fw4.pdf


Hopding (approved):
This will be handy. Thanks @btecu!

PR #696, opened 2020-12-09
Do not add border when none is defined or set to transparent

### LINKED ISSUE #682: Border default
If a `TextField` has a background color but no border (transparent), the library adds a black border of value one by default:
https://github.com/Hopding/pdf-lib/blob/b4d9731ed6beb63d889ce996e13fbebbaf0fb118/src/api/form/appearances.ts#L446

So a field with a white background color end up looking like this:
image

Instead they should look like this:
image

btecu:
The rest of the `AppearanceProvider` are already not defaulting to a border.

The issue happens when a border color is set on an element such as a `TextField`. When the library fills the form, it also looks for a border and if it's not defined or set to transparent, it adds one with the default of `1` width and color `black`.

Closes #682.


Hopding (approved):
Thanks @btecu!

PR #700, opened 2020-12-11
Add image stamping for TextField elements

btecu:
Allows stamping an image on a `TextField` element in addition to `Button` elements.
Additionally, it now handles alignment, defaulting to `Center` for `Button` and using `TextAlignment` for `TextField`.


Hopding (changes requested):
@btecu Good idea! A couple things before merging:
* Would you please update one of the existing integration tests with form field to exercise this new method on `TextField`?
* Can you create a new enum called `ImageAlignment` rather than using the `TextAlignment` enum? Their values happen to be the same right now, but I'd prefer to keep them separated.

btecu:
@Hopding feedback addressed!
I've modified one of the test PDFs and replaced one of the buttons used for image stamping with a text field.

Hopding (approved):
Looks great! Thanks again @btecu. I'll tweak a few things and then this feature will go out in the next release of pdf-lib.

PR #716, opened 2020-12-20
Form Flatten

### LINKED ISSUE #554: [Feature Request] API to Flatten PDF Forms
I haven't been able to find any way to flatten a PDF form, and I'm integrating with an API (Lob) that can't / won't handle non-flattened forms (it won't render the content of any acroFields). I've tried making the fields read-only, and I've set NeedAppearances to `true` but that doesn't cut it for this one use-case.  

Is there a way, even if complicated/hacky to flatten a PDF form with `pdf-lib`?

### LINKED ISSUE #613: [Feature Request] PDF flatten
Hi @Hopding,
Thanks you for your wonderful lib! 
I
One feature I am missing in it though is the ability to flatten form fields, similar to how you can do it in pdftk for example
https://www.pdflabs.com/docs/pdftk-man-page/#dest-output-flatten.
That would allow to merge pdfs with fields having same name but different value, which is currently not possible using pdf-lib.

Thanks !

btecu:
Implements form flattening. Closes #554.


src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+   * of the pages content stream. All form fields and annotations associated are removed.
+   * 
+   * For example:
+   * ```js
+   * const form = pdfDoc.getForm();
+   * form.flatten();
+   * ```
+   */
+  flatten() {
+    this.updateFieldAppearances();
```
Hopding, line 524: In most cases, users will want the field appearances to be automatically updated. But this won't _always_ be the case, so we should make sure users can opt out of this behavior. The best way to handle this is probably to add a `updateFieldAppearances` option to this method, similar to `pdfDoc.save(...)`: https://github.com/Hopding/pdf-lib/blob/b169e2fda05beacfe722ed1b3e73e4ec9a59a589/src/api/PDFDocument.ts#L1223

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+
+    const fields = this.getFields();
+    const pages = this.doc.getPages();
+
+    for (const field of fields) {
```
Hopding, line 529: This syntax is nice and I find it much more readable than low-level indexing loops. However, the decision to use optimized index-loops in the pdf-lib codebase has already been made (for performance reasons). So please convert this loop to use the following syntax: https://github.com/Hopding/pdf-lib/blob/b169e2fda05beacfe722ed1b3e73e4ec9a59a589/src/api/PDFDocument.ts#L1168

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+      const widgets = field.acroField.getWidgets();
+
+      for (const widget of widgets) {
```
Hopding, line 532: Please see previous comment about index-loops.

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+        const pageRef = widget.P();
+        const page = pages.find(x => x.ref === pageRef);
```
Hopding, line 534: We should assert that we did, in fact, find a match. This will also remove the need for non-null assertions later in the method.
```suggestion
        const page = pages.find(x => x.ref === pageRef);
        if (!page) throw new Error(`Failed to find page: ${pageRef}`);
```

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+            const value = field.acroField.getValue();
+            ref = ref.get(value) as PDFRef;
+          } else if (field instanceof PDFRadioGroup) {
+            const value = field.acroField.getValue();
+
+            ref = ref.has(value)
+              ? ref.get(value) as PDFRef
+              : ref.get(PDFName.of('Off')) as PDFRef;
+          }
+        }
```
Hopding, line 548: I think we can streamline the logic a bit and remove some type assumptions:
```suggestion
        let refOrDict = widget.getNormalAppearance();

        if (
          refOrDict instanceof PDFDict &&
          (field instanceof PDFCheckBox || field instanceof PDFRadioGroup)
        ) {
          const value = field.acroField.getValue();
          const ref = refOrDict.get(value) ?? refOrDict.get(PDFName.of('Off'));
          if (ref instanceof PDFRef) refOrDict = ref;
        }

        if (!(refOrDict instanceof PDFRef)) {
          throw new Error(`Failed to extract appearance ref`);
        }
```

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+
+        const xObjectKey = addRandomSuffix('FlatWidget', 10);
+        page!.node.setXObject(PDFName.of(xObjectKey), ref as PDFRef);
```
Hopding, line 551: With the previous changes we can remove some assumptions from this line.
```suggestion
        page.node.setXObject(PDFName.of(xObjectKey), refOrDict);
```

src/api/form/PDFForm.ts
```diff
@@ -499,6 +508,68 @@ export default class PDFForm {
...
+
+        const operators = [
+          pushGraphicsState(),
+          translate(rectangle.x, rectangle.y),
+          rotateRadians(toRadians(rotation)),
+          drawObject(xObjectKey),
+          popGraphicsState(),
+        ].filter(Boolean) as PDFOperator[];
+    
+        page!.pushOperators(...operators);
```
Hopding, line 565:
```suggestion
        page.pushOperators(...operators);
```

src/core/annotation/PDFAnnotation.ts
```diff
@@ -56,6 +56,11 @@ class PDFAnnotation {
     return AP;
   }
 
+  getNormalAppearance() : PDFRef | PDFDict {
+    const AP = this.ensureAP();
+    return AP.get(PDFName.of('N')) as PDFRef | PDFDict;
```
Hopding, line 61: It's best to avoid `as` casts whenever possible (at least in the context of the pdf-lib codebase):
```suggestion
    const N = AP.get(PDFName.of('N'));
    if (N instanceof PDFRef || N instanceof PDFDict) return N;
    throw new Error(`Unexpected N type: ${N?.constructor.name}`);
```

src/core/objects/PDFArray.ts
```diff
@@ -37,6 +37,10 @@ class PDFArray extends PDFObject {
     this.array.splice(index, 0, object);
   }
 
+  indexOf(object: PDFObject): number {
+    return this.array.indexOf(object);
+  }
```
Hopding, line 42: Can we alter this so the "no-match" case is indicated by an `undefined` value? This is more typesafe and will ensure that the caller doesn't forget to handle it.
```suggestion
  indexOf(object: PDFObject): number | undefined {
    const index = this.array.indexOf(object);
    return index < 0 ? index : undefined;
  }
```

Hopding (changes requested):
@btecu Wow... this is awesome! Form flattening is a much wanted feature that will help a lot of users. It's probably the biggest missing form-filling feature. Thanks for working on this! I've left a few pieces of feedback for you.

apps/deno/tests/test18.ts
```diff
@@ -0,0 +1,27 @@
+import { Assets } from '../index.ts';
```
Hopding, line 1: Can we update this test to include all the field types? I'd like to be able to verify that the form flattening feature works for all of them.

Also, it would be nice to verify that once a form is flattened, it's filled "fields" (or, more precisely, their appearances) can be copied into other documents via normal page copying. So perhaps this test could load the documents used in tests 14-17, fill some fields in each, flatten them, and then copy a subset of the pages from each of them into a new document.

Hopding (changes requested):

Hopding (approved):
Looks great. Thanks again @btecu!

chislin:
When running `form.flatten()` on PDF saved from result of executing script from [this](https://pdf-lib.js.org/#create-form) example, I get this error:
```
 Error: Failed to find page undefined for element favorite.superhero
    at _loop_1 (/node_modules/pdf-lib/cjs/api/form/PDFForm.js:463:27)
    at PDFForm.flatten (/node_modules/pdf-lib/cjs/api/form/PDFForm.js:492:17)
```

Hopding:
Looks like this is because the form flattening implementation relies on an optional widget field:

image

`pdf-lib` could set this field when adding widgets to pages. But it does seem unfortunate that form flattening would require the presence of an optional field.

btecu:
@chislin are you able to attach the document with the issue?

chislin:
@btecu attached one
[form_on_3rd.pdf](https://github.com/Hopding/pdf-lib/files/5741532/form_on_3rd.pdf)

btecu:
@chislin can you create an issue?

PR #720, opened 2020-12-23
Check all radio buttons in a group when deciding if the appearance needs to be updated

### LINKED ISSUE #678: Radio button overlap
When filling any radio group, the actual symbol on the page gets overwritten with a radio button. The page might have it look as a checkbox - which PDF supports.
In the best cases you end up with something like this:
image

However that doesn't always happens. It turns out that there is some internal widget ordering and based on that and the selected value it might overlap the radio button or not:
image

btecu:
Fixes #678.

### DIFF
--- a/src/api/form/PDFRadioGroup.ts
+++ b/src/api/form/PDFRadioGroup.ts
@@ -402,10 +402,13 @@ export default class PDFRadioGroup extends PDFField {
       const widget = widgets[idx];
       const value = this.acroField.getValue();
       const normal = widget.getAppearances()?.normal;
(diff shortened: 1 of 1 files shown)


Hopding (approved):
Looks good. Thanks @btecu!

---

Btw, it looks like this same issue might also apply to checkboxes? https://github.com/Hopding/pdf-lib/blob/654334b24a80327f7ac25f3d3791ec53cdca30f7/src/api/form/PDFCheckBox.ts#L201

Did you find those behaved differently, or have you not tested their behavior?

PR #723, opened 2020-12-28
Fallback to finding the page using the reference

### LINKED ISSUE #722: Unable to flatten PDF if widget doesn't have P reference
When running `form.flatten()` on some PDFs, like from this [this](https://pdf-lib.js.org/#create-form) example, I get this error:
```
 Error: Failed to find page undefined for element favorite.superhero
    at _loop_1 (/node_modules/pdf-lib/cjs/api/form/PDFForm.js:463:27)
    at PDFForm.flatten (/node_modules/pdf-lib/cjs/api/form/PDFForm.js:492:17)
```

Here is an [example](https://github.com/Hopding/pdf-lib/files/5741532/form_on_3rd.pdf) of PDF that will produce this error and where form isn't on 1st page. (Originally posted [here](https://github.com/Hopding/pdf-lib/pull/716#issuecomment-751115024))

btecu:
Fixes #722.

### DIFF
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -1265,18 +1265,18 @@ export default class PDFDocument {
     return dataUri ? `data:application/pdf;base64,${base64}` : base64;
   }
 
-  findPageForRef(ref: PDFRef): PDFPage {
-    let pages = this.getPages();
+  findPageForAnnotationRef(ref: PDFRef): PDFPage | undefined {
+    const pages = this.getPages();
     for (let idx = 0, len = pages.length; idx < len; idx++) {
       const page = pages[idx];
       const annotations = page.node.Annots();
-  
+
       if (annotations?.indexOf(ref) !== undefined) {
         return page;
       }
     }
 
-    throw new Error(`Could not find page for PDFRef ${ref}`);
+    return undefined;
   }
 
   private async embedAll(embeddables: Embeddable[]): Promise<void> {
--- a/src/api/PDFDocument.ts
+++ b/src/api/PDFDocument.ts
@@ -46,6 +46,7 @@ import {
   SetTitleOptions,
 } from 'src/api/PDFDocumentOptions';
 import PDFObject from 'src/core/objects/PDFObject';
+import PDFRef from 'src/core/objects/PDFRef';
 import { Fontkit } from 'src/types/fontkit';
 import { TransformationMatrix } from 'src/types/matrix';
 import {
@@ -1262,6 +1263,20 @@ export default class PDFDocument {
     const bytes = await this.save(otherOptions);
     const base64 = encodeToBase64(bytes);
     return dataUri ? `data:application/pdf;base64,${base64}` : base64;
+  }
+
+  findPageForRef(ref: PDFRef): PDFPage {
+    let pages = this.getPages();
+    for (let idx = 0, len = pages.length; idx < len; idx++) {
+      const page = pages[idx];
+      const annotations = page.node.Annots();
+  
+      if (annotations?.indexOf(ref) !== undefined) {
+        return page;
+      }
+    }
+
+    throw new Error(`Could not find page for PDFRef ${ref}`);
   }
 
   private async embedAll(embeddables: Embeddable[]): Promise<void> {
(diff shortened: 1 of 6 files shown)


Hopding:
Please see https://github.com/Hopding/pdf-lib/issues/722#issuecomment-753511561.

btecu:
Bonus: I've added `form.removeField()` since it seems to be a feature other people have been asking for.

src/api/PDFDocument.ts
```diff
@@ -1264,6 +1265,20 @@ export default class PDFDocument {
     return dataUri ? `data:application/pdf;base64,${base64}` : base64;
   }
 
+  findPageForRef(ref: PDFRef): PDFPage {
```
Hopding, line 1268: Can we rename this so that it's clear the ref should be that of an annotation?
```suggestion
  findPageForAnnotationRef(ref: PDFRef): PDFPage {
```

src/api/PDFDocument.ts
```diff
@@ -1264,6 +1265,20 @@ export default class PDFDocument {
...
+    for (let idx = 0, len = pages.length; idx < len; idx++) {
+      const page = pages[idx];
+      const annotations = page.node.Annots();
+
+      if (annotations?.indexOf(ref) !== undefined) {
+        return page;
+      }
+    }
+
+    throw new Error(`Could not find page for PDFRef ${ref}`);
```
Hopding, line 1279: I'd like to avoid throwing an error here in order to keep this method as flexible as possible. It may be used in scenarios where the caller isn't sure whether a ref lookup will succeed or not.
```suggestion
    return undefined;
```

src/api/form/PDFForm.ts
```diff
@@ -591,11 +590,25 @@ export default class PDFForm {
         page.pushOperators(...operators);
       }
 
-      this.acroForm.removeField(field.ref);
-      this.doc.context.delete(field.ref);
+      this.removeField(field);
```
Hopding, line 593: I wonder if we should also remove the widgets themselves from their page's `Annots` array after flattening them? They don't seem to impact how readers render the document, but they are dead weight. Not strictly necessary though.

src/core/PDFContext.ts
```diff
@@ -163,6 +163,16 @@ class PDFContext {
     throw new UnexpectedObjectTypeError(types, result);
   }
 
+  getObjectRef(pdfObject: PDFObject): PDFRef {
+    for (const [ref, object] of this.indirectObjects.entries()) {
+      if (object === pdfObject) {
+        return ref;
+      }
+    }
```
Hopding, line 171: I try to avoid changing the compiler config unless it's strictly necessary. Doing so can cause unexpected issues for library users. If we use `Array.from` we can implement this logic without changing the config.
```suggestion
    const entries = Array.from(this.indirectObjects.entries());
    for (let idx = 0, len = entries.length; idx < len; idx++) {
      const [ref, object] = entries[idx];
      if (object === pdfObject) {
        return ref;
      }
    }
```

src/core/PDFContext.ts
```diff
@@ -163,6 +163,16 @@ class PDFContext {
...
+
+    throw new Error('Could not find PDFRef for PDFObject');
```
Hopding, line 173: This method seems likely to be used in cases where the `PDFObject` isn't yet registered in the context.
```suggestion
    return undefined;
```

tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "tsBuildInfoFile": "./tsBuildInfo.json",
     "removeComments": false,
     "importHelpers": true,
-    "downlevelIteration": false,
+    "downlevelIteration": true,
```
Hopding, line 18: Should be able to avoid this config change per my previous suggestion.
```suggestion
    "downlevelIteration": false,
```

Hopding (changes requested):
Looks great overall! Thanks @btecu. 

Request a few (hopefully small) changes. Then this will be ready to go 👍

btecu:
@Hopding feedback addressed.

Hopding (approved):
Much appreciated. Thanks @btecu! I'm going to update the integration tests to cover all this logic and check some new scenarios for form flattening. Then I'll merge all this to `master`.

PR #724, opened 2020-12-30
Fix checkbox appearance check

btecu:


src/api/form/PDFForm.ts
```diff
@@ -579,6 +579,9 @@ export default class PDFForm {
         const ap = widget.getAppearanceCharacteristics();
         const rectangle = widget.getRectangle();
         const rotation = degrees(ap?.getRotation() ?? 0);
+        if (rectangle.height < 0) {
+          rectangle.y += rectangle.height;
+        }
```
btecu, line 584: Should handle negative `y` as well:
```js
          rectangle.y = rectangle.y < 0
            ? rectangle.y - rectangle.height
            : rectangle.y + rectangle.height;
```

Hopding:
@btecu apologies for my late response. I've been very busy these past few months taking care of some personal stuff and starting my new job at GitHub 🙂 

Is this PR still valid? If so, you'll need to either update this PR (via a force push) or open a new PR with your changes. The tl;dr is that I recently had to do a fairly extensive git history rewrite that caused a bunch of conflicts. Please read https://github.com/Hopding/pdf-lib#git-history-rewrite for details. Apologies for the hassle!

btecu:
@Hopding I pushed to the branch but it doesn't seem to be updated here: https://github.com/Hopding/pdf-lib/compare/master...btecu:radiogroup

Also, it's been a while and I don't remember exactly which edge cases the changes we handling.

Hopding:
Hmm. I'm able to view the diff without trouble, but GitHub still seems to think there are conflicting files for some reason.

Hopding:
After reviewing the diff, I'm not sure I understand what scenarios this change solves for. As far as I can tell the current implementation seems to match the spec. Can you provide any additional context/details? (I'm guessing not since it's been so long).

btecu:
I'll close this for now.

PR #726, opened 2020-12-30
Invalid non-unicode glyphs width calculation on an embedded font

zapkub:
My colleagues and I try to use `pdf-lib` to render documents in the Thai language and it does not render properly

Here is an output of 2 TypeType font which the left one has an extra space follow the first column of the word which only occur for some Thai language support font
image

This is an expected result
image

after investigating the problem we found missing glyphs that have been excluded from the embedded glyphs array

image

These glyphs (without non-Unicode glyphs) somehow they will be calculated the width by `computeWidths` then embedded into PDF file.

### For instance
```
น้ำ
```
This word includes 3 Unicode characters which will be substituted into 2  Unicode characters and 1 Non-Unicode character depend on the Layout common and ligature rule of the font.

to solve the problem we tried to calculate all of the glyphs in the font by using `_glyphs` in the font instance which seems likely to be private property.

The question is Do we need to also change `fontkit` package to support all of the glyphs without looping over the character set array?

and I also cannot find out why this issue only happens in some kind of fonts


Hopding:
Hello @zapkub!

Thanks for working on this. Glad you were able to find the cause of the issue. Looks like it probably took a fair amount of work!

You are correct that `_glyphs` is a private property. As such, I would like to avoid accessing it from `pdf-lib`. I'm no `fontkit` expert, so I'm not sure if this is a bug in the implementation of `font.characterSet` or if a new API needs to be introduced to `fontkit`. 

`@pdf-lib/fontkit` is a fork of [`fontkit`](https://github.com/foliojs/fontkit). The only reason this fork exists is to make `fontkit` builds compatible with the `pdf-lib` build pipeline. It doesn't have new features or anything. So if you make a bug fix or introduce a new API, it would be best to get that into [`fontkit`](https://github.com/foliojs/fontkit) itself and once it's merged pull it into `@pdf-lib/fontkit`. If the authors of `fontkit` are unwilling to accept your changes for whatever reason, then merging it directly into `@pdf-lib/fontkit` is not out of the question. But we'll cross that bridge when we come to it.

One other thing I should mention is that `@pdf-lib/fontkit` is currently 21 commits behind `fontkit`. These commits do need to be merged into `@pdf-lib/fontkit`, I just haven't had the time to do it yet. It's possible they already contain a fix for this issue. But I haven't checked myself.

zapkub:
@Hopding I agree. I will try to fix this issue in `fontkit` ( also I will try to use latest version of fontkit too )

dsohscience37:
> @Hopding I agree. I will try to fix this issue in `fontkit` ( also I will try to use latest version of fontkit too )

@zapkub Did you end up resolving this issue?

zapkub:
> > @Hopding I agree. I will try to fix this issue in `fontkit` ( also I will try to use latest version of fontkit too )
> 
> @zapkub Did you end up resolving this issue?

after digging an issues and during we are in the edge of production release

Me and my team ended up with a directly forked and monkey patching `fontkit` and use it as internal NPM package

will ping here again if we ready to polishing our patch

Hopding:
@zapkub any updates on this? Would love to get your fix into pdf-lib.

Hopding:
Closing this for now. Added it to the roadmap in https://github.com/Hopding/pdf-lib/discussions/998.

PR #730, opened 2021-01-02
Support hidden form fields

jbeuckm:


src/api/form/PDFTextField.ts
```diff
@@ -669,6 +669,7 @@ export default class PDFTextField extends PDFField {
       borderColor: options.borderColor,
       borderWidth: options.borderWidth ?? 0,
       rotate: options.rotate ?? degrees(0),
+      hidden: options.hidden,
```
Hopding, line 672: Would you please add this to all the different field types? (`PDFButton`, `PDFCheckBox`, `PDFDropdown`, `PDFForm`, `PDFOptionList`, `PDFRadioGroup`, `PDFTextField`)

Hopding (changes requested):
Thanks @jbeuckm! This will be a nice addition to the library.

In addition to updating the other field types, would you please update one of the integration tests to add a hidden field? That way this feature doesn't regress down the line.

Also, I'm curious what your use case for a hidden field is, if you don't mind sharing? I've never had occasion to use them myself. Just wondering what purpose they would serve.

jbeuckm:
> Thanks @jbeuckm! This will be a nice addition to the library.
> 
> In addition to updating the other field types, would you please update one of the integration tests to add a hidden field? That way this feature doesn't regress down the line.
> 
> Also, I'm curious what your use case for a hidden field is, if you don't mind sharing? I've never had occasion to use them myself. Just wondering what purpose they would serve.

Yes - I was just being lazy. I added `hidden` to the other fields and tested the flag in the TextField spec. See what you think.

I wanted a non-obvious way to mark a PDF so that I could check that it was (probably) the same original file later. I'm interested in hearing other ways to do the same. If there is a clear path to doing this with an encrypted certificate, I might help work on it.

src/api/form/PDFTextField.ts
jbeuckm: 252b71b

Hopding (approved):

Issue #207, opened 2019-09-29
File size increased hugely after embeded font through fontkit


Pomelool:
I have a pdf that is 600KB, after embeded a font (ttf file of 20MB), and drawed one line with this font, my generated pdf becomes 9mb. Is this a normal behavior? Or am I doing something wrong?

Is there a way to subsetting a font instead of embedding the whole font?

Hopding:
Hello @Pomelool! This is normal. `pdf-lib` does not subset embedded fonts by default. Instead, it embeds the entire font into the document. However, there is an option to use subsetting when you embed the font. It is [an option](https://pdf-lib.js.org/docs/api/interfaces/embedfontoptions) you must pass to [`PDFDocument.embedFont`](https://pdf-lib.js.org/docs/api/classes/pdfdocument#embedfont). For example:
```
const ubuntuBytes = ...
const ubuntuFont = await pdfDoc.embedFont(ubuntuBytes, { subset: true })
```
And here are some other examples: https://github.com/Hopding/pdf-lib/blob/7e06c40c0f19a0008baa885a7740f8968866e2f0/apps/node/tests/test1.ts#L168-L228

It's important to note that subsetting does not work for all fonts. This is why the feature is not enabled by default. The subsetting logic resides in the `@pdf-lib/fontkit` module, which is a fork of https://github.com/foliojs/fontkit. If the subsetting bug in `fontkit` is ever fixed, then I can pull it into `pdf-lib` and perhaps enable subsetting by default.

I hope this helps. Please let me know if you have any additional questions!

vlio20:
Thanks @Hopding . I suggest that you will add this to your documentation in the README.md file

Hopding:
That's a good idea. It would make a nice addition to the https://github.com/Hopding/pdf-lib#fonts-and-unicode section. Happy to accept PRs for this!

vlio20:
Sure I will happy to that!

vlio20:
@Hopding , another question if I may. Is it possible to embed multiple fonts (to increase the chance of successful conversion)?

philfreo:
> The subsetting logic resides in the @pdf-lib/fontkit module, which is a fork of https://github.com/foliojs/fontkit. If the subsetting bug in fontkit is ever fixed, then I can pull it into pdf-lib and perhaps enable subsetting by default.

Has the bug you referenced been fixed yet?
If not, is there a specific GitHub Issue you can share that I can track upstream?

[...]

PR #731, opened 2021-01-02
adding information about font subset

vlio20:
adding information about font subset as discussed in https://github.com/Hopding/pdf-lib/issues/207#


Hopding (approved):
Thanks @vlio20!

PR #742, opened 2021-01-09
Update README.md

lucashmorais:
I thought someone could like to see efforts to produce a free alternative to ILovePDF / SmallPDF based on pdf-lib, so maybe we could include [PDF Shelter](https://pdfshelter.com) in the README.

Cheers!
Lucas


Hopding (approved):
Very cool @lucashmorais!

Issue #744, opened 2021-01-10
Add support for duplex viewer preference


mcshaz:
Similar to the issue for dompdf at https://github.com/dompdf/dompdf/issues/2046 - would it be possible to add the ability to create and set values for a viewer preferences entry into the document's catalog, as described at https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=370 (page 362 of PDF spec 1.7 under 12.2 Viewer Preferences).

Hopding:
Support for this added in v1.16.0

PR #745, opened 2021-01-11
more ViewerPreferences

mcshaz:
address issue #744 and complete the rest of the spec for ViewerPreferences (with the exception of all key-value pairs for which the spec states 'Most conforming readers disregard it.'). includes documentation and unit tests.

potential further tests not included (but I can write if you tell me your thoughts & where you would like them):
1. Test that during serialization the document catalog is the root entry in the trailer.
2. Test that serialization and then deserialization retains all the metadata and also viewerPreferences.

Also address #747 (minor). If you want to support back to ES5, I can alter the Rollup workflow (using babel) so that *both* a modern and ES5 version of the code is produced, but as per the comments in the issue, I think this should actually be the responsibility of the consuming code.

P.S. I tried a rebase but failed - If you would like to merge this code, can you please squash everything to a single commit during the merge- thanks.


Hopding:
Hello @mcshaz! It would be awesome to have setters and getters for the ViewerPreferences. Thanks for working on this.

Over all the code looks good, but there are some spots that will need to be tweaked to match the patterns & conventions of this codebase. I'll leave some specific review comments in the next few days when I have some time. Though some things I'll just polish up myself after merging.

Regarding the rollup and tsconfig changes: If either of those things need to be changed, that should happen in a dedicated PR. I try not to fix code/features with config and compiler changes when possible. For this PR, please implement the logic under the current rollup and tsconfig setup. Please hold off for now on opening a separate PR for rollup/tsconfig changes, those things can impact consumers of the library in unexpected ways and may result in unintentional breaking changes, so they need to be considered very carefully.

mcshaz:
OK I'll remove the target directive - I looked more closely at the TS spec and saw that target might not be a good one to play with as arrow functions will no longer be transpiled to function statements.

However, developing in visual studio code, a significant number of files within this library use functions and objects which are not part of the ES5 spec (listed in #747) and are therefore displayed as warnings within the debugger. What about including a lib directive to ES2016. The lib directive is described at https://www.typescriptlang.org/tsconfig#lib-config - obviously I'll put this as a separate pull request.

mcshaz:
Another question to think about is how to get and set the array pairs to the PrintPageRange. currently it is an array of numbers and the setter checks the array is an even length. Therefore printing pages 1,3 and 5 to 7 would be `[1, 1, 3, 3, 5, 7]`. However it might be better for readability and also TypeScript type checking to be an array of tuples `[number, number][]` such that the values associated with the get & set method for the same page range is denoted by [[1, 1], [3, 3], [5, 7]]`.

Let me know your thoughts.

src/core/interactive/ViewerPreferences.ts
```diff
@@ -18,20 +45,353 @@ class ViewerPreferences {
...
+  protected FindNumber(key: NumberViewerPrefKey): PDFNumber | undefined {
+    const returnObj = this.dict.lookup(PDFName.of(key));
+    if (returnObj instanceof PDFNumber) return returnObj;
+    return undefined;
+  }
+
+  /*
+   * Gets the value of the flag specifying whether to hide the conforming reader’s tool bars when the document is active.
+   * Default value if not specified: false.
+   */
```
Hopding, line 75: These comments are good, but please tweak them a bit to confirm to the comment structure found in other parts of the pdf-lib codebase. Comment structure consistency is as important as code consistency in pdf-lib (even more so for end-users). 

Here's an example of the comment structure for boolean methods: https://github.com/Hopding/pdf-lib/blob/4be96eb269910bbc6d3ddcd48f2a6382742e179b/src/api/form/PDFDropdown.ts#L266-L279

This comment can be rewritten as follows to adhere to this structure:

```suggestion
  /**
   * Returns `true` if PDF readers should hide toolbar menus when displaying
   * this document.
   * @returns Whether or not toolbar menus should be hidden.
   */
```

The code example can be omitted from the comment since this method is part of the `src/core` module rather than the `src/api` module. Though it wouldn't hurt to include anyways. But it's not required.

It would be time consuming to explain in detail exactly what the comment structure is. But I trust the codebase provides sufficient examples to make it clear. The main thing is to keep the wording, grammar, flow, and voice of the comments similar. However, some specific things to watch out for are:
* Use doc-comments `/** ... */` (as opposed to `/* ... */`), unless the comment is a developer note.
* Comments should not exceed the 80 character line limit. Unfortunately the linter and formatter don't check this, so you'll need to do it yourself (VSCode can display a vertical marker at the 80 character point that makes this easy).
* Return types should be documented with an `@returns` tag and an appropriate description.
* Parameter types should be documented with an `@param` tag and an appropriate description.

This note applies to several of the following methods as well, though I won't leave comments on all of them.

tests/api/PDFDocument.spec.ts
```diff
@@ -303,15 +303,65 @@ describe(`PDFDocument`, () => {
...
+  describe(`viewerPreferences`, () => {
     it(`defaults to an undefined ViewerPreferences dict`, async () => {
       const pdfDoc = await PDFDocument.create();
 
       expect(
         pdfDoc.catalog.lookupMaybe(PDFName.of('ViewerPreferences'), PDFDict),
       ).toBeUndefined();
     });
+    it(`can get/set HideToolbar, HideMenubar, HideWindowUI, FitWindow, CenterWindow, DisplayDocTitle, NonFullScreenPageMode, Direction, PrintScaling, Duplex, PickTrayByPDFSize, PrintPageRange, NumCopies from a new document`, async () => {
+      const pdfDoc = await PDFDocument.create();
```
Hopding, line 315: This test is good. Please add a second test that verifies the viewer preferences can be read from an existing PDF document (with `PDFDocument.load(...)`) that already has viewer preferences defined. You'll need to create a new document and adjust all of its viewer preferences in Adobe Acrobat (or another PDF editor that's capable of changing all these properties). Be sure to commit the PDF to the `assets/pdfs` directory as well.

Hopding (changes requested):
Thanks again for working on this @mcshaz! I've gone through the PR and made some change requests. Once they're addressed this should be good to merge and release. Let me know if you have any questions!

src/core/interactive/ViewerPreferences.ts
mcshaz: I have attempted to alter the comments in the code, as per this conversation. The use of string `enums` does not seem to be used elsewhere in the library, so I have no great proforma. there is still an inconsistency with the possible values being listed in the description for the setters and in the `@returns` for the getters. Please let me know which style you prefer.

Hopding: @mcshaz the new comments (and all your other changes) look great! Thanks for addressing my review comments so promptly.

The preferred way to document enum values is via doc comments on the enum value definitions themselves. For example:
```ts
export enum NonFullScreenPageMode {
  /** 
   * After exiting full-screen mode neither the document outline nor thumbnail 
   * images should be visible. 
   */
  UseNone = 'UseNone',

  /** After exiting full-screen mode the document outline should be visible. */
  UseOutlines = 'UseOutlines',

  /** After exiting full-screen mode the thumbnail images should be visible. */
  UseThumbs = 'UseThumbs',

  /** 
   * After exiting full-screen mode the optional content group panel should be 
   * visible. 
   */
  UseOC = 'UseOC',
}
```

This removes the need for duplicate information in the comments, especially if the enum is used in multiple methods. The comments will display as follows in the API docs: 

image

mcshaz: Thank you - I have attempted to address all the suggestions. I now also note that table 28 on page 73 of the spec includes entries in the `catalog` dictionary `PageLayout` and `PageMode`. I cannot see why these are not `ViewerPreferences` in their specification, but c'est la vie. I can add the associated getters/setters for these to the PDFCatalog class in the same manner if you would like - simply because it kind of rounds off the ViewerPreferences, and in truth `NonFullScreenPageMode` does not make much sense without  at least PageMode. Note the documentation now has the example:
```js
// set the PageMode, otherwise the subsequent NonFullScreenPageMode has no meaning
pdfDoc.catalog.set(PDFName.of('PageMode'),PDFName.of('FullScreen'))
// set what happens when fullScreen is closed
viewerPrefs.setNonFullScreenPageMode(NonFullScreenPageMode.UseOutlines) 
```

Hopding: I'd be happy to accept a PR for adding those setters/getters to the `PDFCatalog` class 👍

Hopding (approved):
Thanks again for all your work on this @mcshaz! These new setters/getters will go out in the next release of pdf-lib  🎉

PR #750, opened 2021-01-13
Cleanup all annotations

btecu:
Most viewers, including `PDF.js` and `Preview` do not seem to mind the annotations but it was preventing `Adobe Acrobat` printing a flattened document:

### DIFF
--- a/src/api/form/PDFForm.ts
+++ b/src/api/form/PDFForm.ts
@@ -1,4 +1,5 @@
 import PDFDocument from 'src/api/PDFDocument';
+import PDFPage from 'src/api/PDFPage';
 import PDFField from 'src/api/form/PDFField';
 import PDFButton from 'src/api/form/PDFButton';
 import PDFCheckBox from 'src/api/form/PDFCheckBox';
@@ -38,6 +39,7 @@ import {
   PDFRef,
   createPDFAcroFields,
   PDFName,
+  PDFWidgetAnnotation,
 } from 'src/core';
 import { addRandomSuffix, assertIs, Cache, assertOrUndefined } from 'src/utils';
 
@@ -538,7 +540,6 @@ export default class PDFForm {
     }
 
     const fields = this.getFields();
-    const pages = this.doc.getPages();
 
     for (let i = 0, lenFields = fields.length; i < lenFields; i++) {
       const field = fields[i];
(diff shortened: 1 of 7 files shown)


weinspach:
The error with the annotations does not only affect printing, "save as" is also not possible in Acrobat Reader and Acrobat Pro with a flattened pdf. In addition, e.g. the validation of the PDF with Datalogic's PDF Checker fails: "Unable to process document due to PDF Error".

src/api/form/PDFForm.ts
```diff
@@ -594,11 +594,15 @@ export default class PDFForm {
...
         page.pushOperators(...operators);
-        page.node.removeAnnot(refOrDict);
       }
 
       this.removeField(field);
     }
+
+    for (let i = 0, len = pages.length; i < len; i++) {
+      const page = pages[i];
+      page.node.clearAnnot();
```
weinspach, line 604: I'm not that familiar with the PDF specification yet, but only widget annotations should be deleted here, right?

btecu: That's what previous code was doing but appears to not have been sufficient.

Is there a way to distinguish the type of the annotations?

weinspach: Maybe the error was caused by the rollover and down appearances that were still in the document for some widgets?

Hopding:
@btecu Can you clarify what is being done here? It looks like this logic would just delete the first annotation from a page after flattening a widget that belonged to it. If that is the intent, can you explain why that would be desirable? There's not necessarily any correlation between the first annotation on a page and the widget being flattened.

Hopding:
Can you provide a sample document that fails to be printed after flattening? Does this only happen to certain documents, or to all documents?

btecu:
@Hopding Notice the loop - is deleting all annotations. As a sample, take a look at the document in the flattening test. It doesn't appear to be a special document but the issue happens with most PDFs that have a form.

Hopding:
Ah, I see the loop now. Deleting all the annotations seems like overkill. I don't think users would expect that flattening a form would result in all pages with fields losing their annotations.

btecu:
Looks like the field annotations are the ones that were causing the issue. Removing those along with the widget annotations seems to have fix the issue.

In addition, I've improved `removeField` so that it cleans up the annotations.

src/api/form/PDFForm.ts
```diff
@@ -612,6 +581,20 @@ export default class PDFForm {
    * ```
    */
   removeField(field: PDFField) {
```
Hopding, line 583: Would you please add some unit tests for this method? This isn't something that can really be verified via the integration tests and it has a fair amount of logic now.

Hopding (changes requested):
@btecu This looks very good. The implementation is nice. I've requested one change and then this should be good to go.

Hopding (approved):
Thanks for working on this @btecu!

Issue #727, opened 2020-12-31
Need way to change font size for text inserted into existing field


duffyd:
* We're using the PDF-LIB for filling PDFs (https://pdf-lib.js.org/docs/api/classes/pdftextfield). It is likely not relevant but we also embed a font for characters outside of the winansi set (Latin)
* On Nov 23rd, someone asked the author (Hopding) if there is a way to change the font size used. He said, "There are not currently any APIs for changing the font size of an existing field. But such an API could certainly be built. Happy to provide some pointers if anybody is interested in making a PR for it."
* This issue is the formal request for pointers :)

btecu:
Look at the code that retrieves the font size:
https://github.com/Hopding/pdf-lib/blob/2b4ee6b097b7231935eac610f0e43a1befa7cac6/src/api/form/appearances.ts#L123-L135

You just have to reverse the process. The `DA` (default appearance) is a string (or HEX) containing the font name, size and a few possible other details.

duffyd:
@btecu thanks for your reply! So just to clarify, should I be setting the **DA** property (ref [here](https://github.com/Hopding/pdf-lib/blob/2b4ee6b097b7231935eac610f0e43a1befa7cac6/src/core/acroform/PDFAcroField.ts#L89)) on the PDF field to a string representing what I want it to look like before doing a **getTextField**?

btecu:
Yes, you can use `getDefaultAppearance`, modify the string to reflect the font size that you want and then use `setDefaultAppearance`: https://github.com/Hopding/pdf-lib/blob/2b4ee6b097b7231935eac610f0e43a1befa7cac6/src/core/acroform/PDFAcroField.ts#L85-L87

duffyd:
@btecu: Great, thanks again!!

Hopding:
@duffyd Just want to mention that what @btecu said is accurate.

Are you planning to make a PR that introduces an API for changing a text field's font size? If so, note that the `DA` property is optional and may not have a font size already defined in it. So you'll need to handle the case of a text field having a `DA` with a font size defined and the case where it does not have a font size defined (and the case of there being no `DA` property at all).

If you're not planning to make a PR, then I'll go ahead and close this issue.

duffyd:
Hi @Hopding, I'm happy to do a PR but just concerned whether you require some tests as well? Just I've never done testcases/etc in JS before. JS coding is fine, though.

Hopding:
[...]

PR #754, opened 2021-01-16
Adding ability to change font size in PDF form fields

duffyd:
#727 Here's the PR. Please let me know if there's anything that needs to be changed. Would be great if this could be released soon so we can start using it in our app 🙂


apps/node/tests/test19.ts
```diff
@@ -0,0 +1,25 @@
+import { Assets } from '..';
```
Hopding, line 1: This is a good test, but please add the logic to one of the existing form filling tests. The integration tests take a lot of time to run, so we try to keep the number of files to a minimum.

Hopding: Also, please be sure to add the same logic to the `deno`, `web`, and `rn` tests. The tests should always be identical across each runtime/environment. Note that you may not be able to actually run the `rn` tests. That's fine, but please still update the file. I'll run it along with the others before releasing the changes.

src/core/acroform/PDFAcroField.ts
```diff
@@ -7,6 +7,20 @@ import PDFNumber from 'src/core/objects/PDFNumber';
 import PDFArray from 'src/core/objects/PDFArray';
 import PDFRef from 'src/core/objects/PDFRef';
 
+const findLastMatch = (value: string, regex: RegExp) => {
```
Hopding, line 10: This function is very similar to the following: https://github.com/Hopding/pdf-lib/blob/4be96eb269910bbc6d3ddcd48f2a6382742e179b/src/api/form/appearances.ts#L111-L121

Please move this function into https://github.com/Hopding/pdf-lib/blob/master/src/utils/strings.ts and remove the function defined in `appearances.ts` to avoid duplicate code and logic.

src/core/acroform/PDFAcroField.ts
```diff
@@ -7,6 +7,20 @@ import PDFNumber from 'src/core/objects/PDFNumber';
...
+  while (position < value.length) {
+    const match = value.substring(position).match(regex);
+    if (!match) return { match: lastMatch, pos: position };
+    lastMatch = match;
+    position += (match.index ?? 0) + match[0].length;
+  }
+  return { match: lastMatch, pos: position };
+};
+
+const tfRegex = /\/([^\0\t\n\f\r\ ]+)[\0\t\n\f\r\ ]*(\d*\.\d+|\d+)?[\0\t\n\f\r\ ]+Tf/;
```
Hopding, line 22: Please add some comments illustrating how this regex matches various inputs. Similar to what was done here: https://github.com/Hopding/pdf-lib/blob/4be96eb269910bbc6d3ddcd48f2a6382742e179b/src/api/form/appearances.ts#L123-L126

src/core/acroform/PDFAcroField.ts
```diff
@@ -96,6 +110,24 @@ class PDFAcroField {
     return DA?.asString();
   }
 
+  setFontSize(fontSize: number) {
+    if (this.DA()) {
+      const da = this.getDefaultAppearance()!;
+      const daMatch = findLastMatch(da, tfRegex);
```
Hopding, line 116: Can we do the following to make the null check safer and avoid the non-null assertion?
```suggestion
  setFontSize(fontSize: number) {
    const da = this.getDefaultAppearance();
    if (da) {
      const daMatch = findLastMatch(da, tfRegex);
```

src/core/acroform/PDFAcroField.ts
```diff
@@ -96,6 +110,24 @@ class PDFAcroField {
     return DA?.asString();
   }
 
+  setFontSize(fontSize: number) {
```
Hopding, line 113: Does this function work for newly created text fields? E.g. https://pdf-lib.js.org/docs/api/classes/pdfform#createtextfield. It's important that form methods work for all types of fields.

Hopding, line 113: Please add wrapper methods to [`PDFTextField`](https://github.com/Hopding/pdf-lib/blob/master/src/api/form/PDFTextField.ts), [`PDFButton`](https://github.com/Hopding/pdf-lib/blob/master/src/api/form/PDFButton.ts), [`PDFDropdown`](https://github.com/Hopding/pdf-lib/blob/master/src/api/form/PDFDropdown.ts), and [`PDFOptionList`](https://github.com/Hopding/pdf-lib/blob/master/src/api/form/PDFOptionList.ts) with the appropriate doc comments. Defining the logic in the low-level `PDFAcroField` API makes sense, but this is functionality that should also be available in the high-level API. Most users should not need to access the `.acroField` property or the low-level core APIs.

Note that the wrapper methods should assert that their `fontSize` parameter is non-negative.

Hopding, line 113: Please add some unit tests for this method to https://github.com/Hopding/pdf-lib/blob/master/tests/core/acroform/PDFAcroField.spec.ts. Parsing logic like this is error prone so it's important to test carefully to avoid regressions and missed edge cases.

Hopding (changes requested):
Thanks for working on this @duffyd! Your approach looks solid and the code is good. I've requested some changes. Once these things are taken care of this can be merged and released.

src/core/acroform/PDFAcroField.ts
duffyd: Yes, I've also implemented integration tests for this. The node one is here: https://github.com/duffyd/pdf-lib/blob/efbc0ee4da21d7cb6a4ed4f61859326148449e8b/apps/node/tests/test17.ts#L110-L113

duffyd:
@Hopgding I've implemented all requested changes and they're in my fork now. Could you please review and let me know if you're happy?

Hopding:
Thanks @duffyd! I'll take a look as soon as I get a chance. Looks like there's a conflict though. Would you please take a look at that?

duffyd:
> Thanks @duffyd! I'll take a look as soon as I get a chance. Looks like there's a conflict though. Would you please take a look at that?

I've fixed the conflict but it's still complaining. Sorry I'm not a git expert. Are you able to resolve the conflict?

duffyd:
Ok, after *reading the documentation* I have now properly resolved the conflict.

Hopding (approved):
Looks great @duffyd! Really appreciate your work on this. It'll go out in the next release of pdf-lib (hopefully tonight or tomorrow).

duffyd:
Thanks!!

PR #817, opened 2021-03-07
switch PDFPageEmbedder from contex.stream to .flateStream to compress page

### LINKED ISSUE #639: Embedded pages are uncompress
Hi,

I'm migrating our pdf generation tool from [`HummusJS`](https://github.com/galkahana/HummusJS) to [`pdf-lib`](https://github.com/Hopding/pdf-lib).

I have to merge two pdf pages into a new pdf. To achieve that I embed the pages into a new pdf document.
The size of the generated file is significantly increased compared to the original one.
When I compress the generated pdf, the file size is closer than what I expect.

I did a test case to show the issue:

```typescript
import {readFileSync, writeFileSync} from 'fs'
import {PDFDocument} from 'pdf-lib'

(async () => {

    const pdfSource = await PDFDocument.load(readFileSync('./Lorem ipsum dolor sit amet.pdf'));
    const embeddedPage = await pdfDoc.embedPage(pdfSource.getPage(0));

    const page = pdfDoc.addPage();
    page.drawPage(embeddedPage);

    writeFileSync('embedded.pdf', await pdfDoc.save());

})();
```

The source file is a simple pdf of 200 KB made with Word : [Lorem ipsum dolor sit amet.pdf](https://github.com/Hopding/pdf-lib/files/5402220/Lorem.ipsum.dolor.sit.amet.pdf)
The file generated by this script is 1.5 MB. It's 7.5x larger : [embedded.pdf](https://github.com/Hopding/pdf-lib/files/5402223/embedded.pdf)

I think that the LZW stream from the source file is uncompressed before being embedded into the new PDF file.

Am I correct?

**EDIT:** I inspected both PDF and I found a FlateDecode stream in the source pdf which is decoded in the destination pdf.
Screenshot 2020-10-20 at 09 48 14

Regards,

Julien

cymen:
I was curious about #639 and spent some time digging into `pdf-lib` and came up with a solution. I realize I'm changing the default from a regular stream to a compressed stream for the embedded PDF page but after some reflection, it seemed like this would be what most/all would want anyway?

Resolves #639

### DIFF
--- a/src/core/embedders/PDFPageEmbedder.ts
+++ b/src/core/embedders/PDFPageEmbedder.ts
@@ -96,7 +96,7 @@ class PDFPageEmbedder {
     const decodedContents = this.decodeContents(Contents);
 
     const { left, bottom, right, top } = this.boundingBox;
-    const xObject = context.stream(decodedContents, {
+    const xObject = context.flateStream(decodedContents, {
       Type: 'XObject',
       Subtype: 'Form',
       FormType: 1,


cymen:
Oh, and I also leant towards just changing from `stream` -> `flateStream` as that is what the `FileEmbedder` is using too (`flateStream` that is).

Hopding:
Hello @cymen! Thanks so much for working on this! Apologies for my late response. I've been very busy these past few months taking care of some personal stuff and starting my new job at GitHub 🙂 

cymen:
@Hopding I've updated my PR to be based off of current master.

Hopding (approved):
Looks great. Thank you @cymen! 🎉

enoh-barbu:
It's not working...

TrungRueta:
Dear @Hopding ,

Can you help take a look this case. Issue still happend, i have 2 files 800kb and 600kb, after merge the final result raised to 29mb. :(

stul-5230:
not work

PR #823, opened 2021-03-12
Add drawSvg feature and fix drawSvgPath inverted y axis

### LINKED ISSUE #822: Reopen #223 : Despite the explainations, the coordinates system really is wrong.
You can check [this](https://jsbin.com/dufefihebo/edit?html,output). You will see that the origin is correctly placed at the bottom-left of the page, but the y axis is not looking upwards as you can see in the example where you need to **remove** pageHeight to get the correct coordinates.

The coordinates in the usual page system should be calculated with: `y => pageHeight - y` and not `y => y - pageHeight`

I'll try to fix.

Sharcoux:
Based on the work of Vaniel, I fixed most of the transform operations so that we can now have a nice and working `drawSvg` feature that should be working for most of svg.

We probably need to fix some stuff for rotations: The current rotation angle should be stored in the inherited props and then applied in the draw functions. The coordinates changes implied by the rotation are already taken into account, so it's just adding the rotation of the object itself. Same goes for skewX and skewY.

This also fixes [#822](https://github.com/Hopding/pdf-lib/issues/822) That I needed to conclude the task.


Sharcoux:
@Hopding any chance to see this being merged? Should I keep working on this lib ? Is it abandoned ? Is it maintained ?

alwins0n:
+1 for this useful PR. from what I understand there is a lot of CSS attributes that make the translation complex, of which the most useful are supported in this PR right? (reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/Presentation ). 

I have been pondering myself if a perfect translation of SVG to pdf vector graphics is (easily) possible since the former seems to be more powerful. I will watch this with interest and hope that the maintainer gets back to this at some point

Sharcoux:
Those interested in this PR can try it out by using the package: [@cantoo/pdf-lib](https://www.npmjs.com/package/@cantoo/pdf-lib)

alwins0n:
Just one thought here: could't your PR be released as its own artifact? the way I see it, it translates svg elements to pdf-native vector elements or be more precise translate to pdf-lib calls which add them.

so it's an add-on that sits on top of pdf-lib somehow (it even delegates all calls internally)

Sharcoux:
Indeed. But how to "add it on"? I don't see what you imagine the kind of api/lib I should expose

alwins0n:
I see several options here from tightly coupled to loosely coupled for a package like pdf-lib-svg or such
- expose a drawSvg function via prototype enhancement on PDFPage OR
- make the package expose a single function drawSvg accepting a pdf-lib PDFPage
both options depend on pdf-lib obviously.

what I would suggest (loosely coupled) is a package that exposes a function accept an svg and returns an object of tag and attributes (which happen to match the pdf-lib ones for the respective tag).

e.g a return object of `foo(svg: string)` could be `{tag: 'text', x: .. y: .., font: ... opacity: ...}` which fits the DrawOption type of the function used for tag (this discrimination is responsibility of the caller)

e.g. 

```
const result = foo(mySvg);
switch (result.tag) {
  case 'rect' => return page.drawRectangle(result);
  case ....
}
```

that way you have a library that converts svgs into pdf native instructions and these interact out-of-the box with pdf-lib.
that library could be used standalone to just extract information from svgs.

just trying to be helpful here and give an idea. don't know if this makes sense

Sharcoux:
@alwins0n @Hopding I can move this to a tertiary package, but I need the change made in `src/api/operations.ts` to be merged before it can work. Check https://github.com/Hopding/pdf-lib/issues/822 for details

Hopding:
Hello @Sharcoux! Thanks so much for working on this! Apologies for my late response. I've been very busy these past few months taking care of some personal stuff and starting my new job at GitHub 🙂 

Sharcoux:
Hey Hopding! Congrats for your job!

I'm glad that you keep working on this lib, because the ecosystem really needs it!

I'll do the rebase and all, but, first, ensure that you are really open to merge it this way. If you plan on a different approach, it might be better to do it right away. Also, this fixes #822 which is probably better, but might create some breaking changes.

Please confirm if you're aligned with this.

Hopding:
@Sharcoux your PR description sounds reasonable, but I can't say for sure without seeing a diff of your changes. And that's difficult to do now that I've pushed the history rewrite 🥲

Sharcoux:
Ok. We have an upcoming fix for the current implementation and we are going to rebase our work on top of master.

By the way, we are currently trying to solve an issue when the svg goes out of the bounds of the pdf. In such a case, the drawing instructions are being ignored. We are going to manually calculate the portion of the svg that fits in the pdf page, but isn't there a built-in way to do this?

Hopding:
> Ok. We have an upcoming fix for the current implementation and we are going to rebase our work on top of master.

Sounds good 👍 

> By the way, we are currently trying to solve an issue when the svg goes out of the bounds of the pdf. In such a case, the drawing instructions are being ignored. We are going to manually calculate the portion of the svg that fits in the pdf page, but isn't there a built-in way to do this?

Not quite sure what you're looking for here. Most (if not all) PDF readers/renderers ignore content drawn outside of a page's bounds.

Sharcoux:
> 
> Not quite sure what you're looking for here. Most (if not all) PDF readers/renderers ignore content drawn outside of a page's bounds.

The issue we met is that svg can have content that cross the viewbox. From what we perceived, drawing such a line in pdf-lib will result in the line not appearing at all. Is it clearer?

Hopding:
Yes, that is more clear. I'm a bit surprised by that behavior. Did you observe it happening in all PDF readers, or just a particular one?

Sharcoux:
No, sorry, I'm getting confused. The issue is that we want to draw only the viewbox in the pdf and not show the part of the svg that is outside of it. But we're on it, and there is nothing you can do. My bad.

geotom:
I really would love to see a drawSVG in pdf.lib. It's a major pain for me. PDFKit has great support via other libraries

Sharcoux:
@geotom we reopened the PR with updated code, don't worry, it will make its way to pdf-lib.

Sharcoux:
Here is the new PR https://github.com/Hopding/pdf-lib/pull/1073

geotom:
Cool. Awaiting this eagerly

PR #884, opened 2021-05-18
PDFBuilder and BDFTable helper class to create PDF files

FlorianPons:
PDF-lib was a great help to DP4CC project. We've created some additional helper class to create PDF files and we would like to share it with the community.


FlorianPons:
Checks failed, we will solve this before merging

PR #924, opened 2021-06-28
Allow custom image key

### LINKED ISSUE #923: Same pdfs have different buffers
I have a code that makes a pdf using this library. Running the code twice gives me two seemingly identical pdfs with the same title, metadata, size, and content. However, when I load and compare the buffers using fs, they appear to have different buffers. Is it possible to make the buffers consistent? I'm trying to write tests and all the tests are failing because I can't just compare the buffers directly.

ChingChang9:
Fixes #923 by adding an optional parameter to `PDFPageDrawImageOptions`

I've ran integration tests in addition to unit tests. They all passed. However, I did not write a test to check whether the custom image key is properly written into the `pool` variable in `PDFName`, since I wasn't able to figure out how to check whether a certain key is in `pool` (calling `Name.of` if the key doesn't exist would just set it on the spot and return it). Looking at [the line where `PDFName.of` is called](https://github.com/Hopding/pdf-lib/blob/d213f921237daca7a5e21fb38eb69d733ff34796/src/api/PDFPage.ts#L977), and the `PDFName.of` method, I think the abscence of this test should be okay(?). If otherwise, I'd need some help

### DIFF
--- a/tests/api/PDFDocument.spec.ts
+++ b/tests/api/PDFDocument.spec.ts
@@ -172,6 +172,49 @@ describe(`PDFDocument`, () => {
     });
   });
 
+  describe(`drawImage() method`, () => {
+    it(`serializes the same value on every save when using a custom image key`, async () => {
+      const imageBuffer = fs.readFileSync('assets/images/mario_emblem.png');
+      const imageKey = 'uniqueImageKey';
+      const pdfDoc1 = await PDFDocument.create({ updateMetadata: false });
+      const pdfDoc2 = await PDFDocument.create({ updateMetadata: false });
+
+      const imageEmbed1 = await pdfDoc1.embedPng(imageBuffer);
+      const imageEmbed2 = await pdfDoc2.embedPng(imageBuffer);
+
+      const page1 = pdfDoc1.addPage([imageEmbed1.width, imageEmbed1.height]);
+      const page2 = pdfDoc2.addPage([imageEmbed2.width, imageEmbed2.height]);
+
+      page1.drawImage(imageEmbed1, { imageKey });
+      page2.drawImage(imageEmbed2, { imageKey });
+
+      const savedDoc1 = await pdfDoc1.save();
+      const savedDoc2 = await pdfDoc2.save();
+
+      expect(savedDoc1).toEqual(savedDoc2);
+    });
+
+    it(`does not serialize the same on save when not using a custom image key`, async () => {
+      const imageBuffer = fs.readFileSync('assets/images/mario_emblem.png');
+      const pdfDoc1 = await PDFDocument.create();
+      const pdfDoc2 = await PDFDocument.create();
+
+      const imageEmbed1 = await pdfDoc1.embedPng(imageBuffer);
+      const imageEmbed2 = await pdfDoc2.embedPng(imageBuffer);
+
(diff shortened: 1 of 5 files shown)


bendersej:
While your fix is great, I believe that a better approach would be to make the key generation deterministic by using `seedrandom`, similarly to what was done as part of this PR: https://github.com/Hopding/pdf-lib/pull/574/files

_Why?_
- Similar convention with one way to do thing everything determinism is expected and required
- No need for the additional image key that provides zero value aside from the deterministic generation (which is very valuable and should definitely be the default behaviour)

ChingChang9:
Thanks for the suggestion. As for the random part, I assume we are just using random prefixes to keep the fonts/images/pages distinct right? If the randomness isn't for security reasons, [would this psuedo-random function on stackoverflow](https://stackoverflow.com/a/19303725/10254049) work? (not uniformly distributed)

bendersej:
_I realise that the PR I mentioned was closed but not merged, my bad!_

> If the randomness isn't for security reasons, would this pseudo-random function on stackoverflow work? (not uniformly distributed)

I believe that it's a perfect candidate!

For good measure you could add a comment to point out that this is not cryptographically secure but not a concern in the intended use-case.

As for the implementation, as mentioned by @Hopding over [here](https://github.com/Hopding/pdf-lib/pull/574#issuecomment-723485108), the best would be to add a `generateRandomNumber` type of `src/utils` and expose it as part of the [`PDFContext`](https://github.com/Hopding/pdf-lib/blob/master/src/core/PDFContext.ts#L46).

This will then nicely fit, with a logic similar to the original PR: https://github.com/Hopding/pdf-lib/pull/574/files#diff-446907be7bbde664b082d8429d463a53099555b54e5d379edaee6220a7d42a52R34, where `rng` is the function called from the context.

_I am not affiliated with PDF-lib, as such, these are only pointers and recommendations, @Hopding is the owner of the repo and has the final word on these decisions_

Hopding:
Hello @ChingChang9! Thanks so much for working on this! Apologies for my late response. I've been very busy these past few months taking care of some personal stuff and starting my new job at GitHub 🙂 

ChingChang9:
No worries. Congratulations on your new job at Github! 😃

I wanted to keep this discussion and pull request, so I force updated my fork and re-committed the same changes. It should be the same as what I originally had now.

tests/api/PDFDocument.spec.ts
```diff
@@ -144,37 +144,41 @@ describe(`PDFDocument`, () => {
...
+      const page2 = pdfDoc2.addPage([imageEmbed2.width, imageEmbed2.height]);
+
+      page1.drawImage(imageEmbed1);
+      page2.drawImage(imageEmbed2);
 
       const savedDoc1 = await pdfDoc1.save();
       const savedDoc2 = await pdfDoc2.save();
 
-      expect(savedDoc1).not.toEqual(savedDoc2);
+      expect(savedDoc1).toEqual(savedDoc2);
```
Hopding, line 181: This seems redundant. Don't think we need this at all anymore?

Hopding (changes requested):
@ChingChang9 Looks good overall. Thanks for working on this! I've just requested a few changes.

Hopding:
Also, I just noticed that this PR appears to still have **387** commits even after the force push. You'll probably need to create a new fork and open a different PR to get rid of them.

PR #912, opened 2021-06-10
Compute multiline font size for multiline inputs

milesnash:
Addresses lack of support for multiline input automatic font sizing, as described in #910 

Ran through the testing as described in CONTRIBUTING.md.  All looked good to me.

Issue #910, opened 2021-06-09
Update `computeFontSize` to support automatic line-wrapping and automatic font size calculation.


mnash-dbg:
As per the comment at https://github.com/Hopding/pdf-lib/blob/master/src/api/text/layout.ts#L144, _"This is hardcoded to make it easier to perform automatic line-wrapping."_.

This greatly reduces the usability of the library for those who need to leverage multiline inputs with a font size of "Auto".

Please consider this issue as a request/placeholder to implement this feature and make its current absence more prominent to consumers of the library.

Hopding:
This has been shipped. Thanks again @milesnash!

PR #983, opened 2021-09-04
Compute multiline font size for multiline inputs

milesnash:
Originally #912.

Addresses lack of support for multiline input automatic font sizing, as described in #910.

Ran through the testing as described in CONTRIBUTING.md. All looked good to me.


milesnash:
Hey @Hopding, just an update to let you know I've re-forked and recreated a branch for this feature 👍

Hopding:
@milesnash thanks for working on this! Would definitely like to close this feature gap 👍 

Can you please:
* Provide some examples/screenshots and explain how this works? 
* Update one or two integration tests to showcase this behavior.
* Add a unit test (using one of the standard fonts) to test this. The most practical way is probably to test it indirectly. E.g. just write unit tests for `layoutMultilineText`.

milesnash:
> @milesnash thanks for working on this! Would definitely like to close this feature gap 👍 
> 
> Can you please:
> * Provide some examples/screenshots and explain how this works? 
> * Update one or two integration tests to showcase this behavior.
> * Add a unit test (using one of the standard fonts) to test this. The most practical way is probably to test it indirectly. E.g. just write unit tests for `layoutMultilineText`.

Sure @Hopding , I'll let you know once I've had a chance to update the PR

milesnash:
Hey @Hopding , apologies for the delay in coming back to you on this, I've been really busy at work of late!

Please see below my responses to your points in your last comment:

## How it works

The inputs to the looping structures within `computeFontSize` are unchanged.

There is an outer while loop that will continue until a return condition is reached or the `fontSize` reaches the max font size.

There is an inner for loop that iterates over each line supplied to the function and determines how much of the available space within the layout bounds of the input will be consumed by the text at the current font size, starting from the minimum font size.  This works by:

1. splitting each line on a space character,
1. reducing on the resultant array and, for each "word":
    1. appending a space character if we are not dealing with the last "word",
    1. calculating its width using the built-in `widthOfTextAtSize` function (as per the previous logic, but applied to portions of the line rather than the whole line at once); and
    1. using the calculated width to update the accumulator which is a tally of the lines we would use at this font size.

After the inner for loop has completed, we will immediately return from the function with the current font size minus 1 if we are not dealing with a multiline input and the lines we have used at the current font size exceeds the lines available.  This ensures we honour the function's existing signature in the case of non-multiline text.  Whether or not the input is multiline is determined by a new, optional flag parameter to the function, `multiline`.

Thereafter the logic is essentially as-is, we calculate whether we have exceeded the available height.  If so, we return the current font size minus 1, if not we increment the font size and the while loop continues.

## Integration tests

I can add some integration tests if absolutely necessary but as I explained in my original PR, I felt that "test15", which is full of multiline inputs, already does an admirable job of that.  To illustrate this, please see below a split-screen of the before/after of that test; the left-hand side being what the test produces on master and the right-hand side being this branch.  I think this handily doubles up as a screenshot/example of the functionality.  For completeness, I have also uploaded the pdfs themselves.

Test15BeforeAfter

[test15-master.pdf](https://github.com/Hopding/pdf-lib/files/7356265/test15-master.pdf)
[test15-compute-multiline-font-size.pdf](https://github.com/Hopding/pdf-lib/files/7356267/test15-compute-multiline-font-size.pdf)

## Unit tests

As you will see I have added unit tests to exercise the functionality indirectly using `layoutMultilineText` as per your suggestion.

Hopding (approved):
@milesnash looks fantastic! Appreciate the detailed explanation and the well-written tests. This will go out in the next release of pdf-lib. Looking forward to it! 🎉

milesnash:
@Hopding awesome thanks, my pleasure to contribute! 🙌

thehamzajunaid:
> It would be very nice if the max font size could be capped at a certain (sane) level to prevent huge small texts like in the example "Gold coins Treasure chests"...

I agree with this. I have a Text Field of fixed height. When I am adding a long piece of text, all of it is getting squeezed to fit in. Similarly when there are only 4,5 words then their size is embarrassingly large. Can someone please tell me what is it that im doing wrong?

PR #859, opened 2021-04-15
fix copy paste typo in radioGroup example

adrianjost:

PR #985, opened 2021-09-06
fix copy paste typo in radioGroup example

adrianjost:
successor of #859


Hopding (approved):
Thanks for fixing this @adrianjost!

Issue #140, opened 2019-07-23
[Feature Request]: PDF should decrease in file size after calling `removePage`


jdbaculard:
Hi

I use PDF-LIB with nodeJS. I've got a PDF File with 2000 pages for 13MB.
I remove 1900 pages (with pdfDoc.removePage(10),pdfDoc.removePage(11),pdfDoc.removePage(12), ...). It's working. it's great but the file size remains the same while only 100 pages remain in the modified file. 

Why ?

Any solution please ? Thanks

Hopding:
Hello @jdbaculard!

The `removePage` method does not delete all of a page's objects from the document. It just removes the reference to the from the page tree (leaving the page's objects intact). The reason for this is that a page can actually be reused. If the objects themselves were deleted, then removing a single page might have unintended side effects elsewhere in the document. And even it it wasn't used multiple times, you may wish to reinsert the page after removing it (for example, [reordering pages](https://github.com/Hopding/pdf-lib/issues/139#issuecomment-513510466)).

That being said, I can certainly see how it would be desirable to actually delete a page's objects from the document, in addition to removing its reference from the page tree. I think it would make sense to add a new method, `PDFDocument.deletePage` that does this.

But, in the meantime, there is a workaround: You can copy the pages you want to keep into a document object. For example, if you want to keep pages 1, 4, and 75:
```js
const originalDoc = await PDFDocument.load(...)
const modifiedDoc = await PDFDocument.create()

const pages = await modifiedDoc.copyPages(originalDoc, [0, 3, 74])
pages.forEach(page => modifiedDoc.addPage(page))

const pdfBytes = await pdfDoc.save()
```

> **NOTE:** This code sample is based on the new v1.0.0 API. This is currently [in beta](https://github.com/Hopding/pdf-lib/issues/134). However, it is stable at this point and will graduate from beta to stable this weekend. So feel free to use it now.

I hope this helps. Please let me know if you have any additional questions!

jdbaculard:
Thank you for your reply.
The copy method is not a good idea for me because each page keeps the logo and each page is 200KB. At the end the total weight of the file is 20MB for 100 pages is much larger than the original file of 2000 pages.

Can I test the deletePage in beta method?

Congratulations for your work.

Hopding:
[...]

PR #986, opened 2021-09-07
Add the ability to copy a pdf

### LINKED ISSUE #951: Corrupted PDF
Hi.

This is an amazing library. Thanks a lot @Hopding. I know you've been inactive for a while but the quality of the code and the support you gave for this during your active time has been absolutely phenomenal. You don't see such fantastic support even for paid products. Good luck whatever you're up to.

My issue is this: I have this PDF file than looks like this:

Screen Shot 2021-07-31 at 9 06 17 PM

But when I open/save it using pdf-lib, it will look like this:

Screen Shot 2021-07-31 at 9 10 36 PM

Has anyone ever had a similar experience?

mohamedsalem401:
closes #951 
Some badly-made PDF files can be automatically repaired by Adobe Reader, but can't be edited by pdf-lib.
Sometimes, working with a copy of these corrupted pdf can fix those issues.

Examples: 
**1- #951**  
https://github.com/Hopding/pdf-lib/files/6912122/test.pdf

### **opening and saving the original file:**
image
### **opening and saving with PDFDocument.load(......, {returnCopy: true})**
image

**2- #140** 
 Having the ability to load a copy directly allows you to remove all of the deleted pages' objects from the document. (thus reducing file size):

### **A file with deleted pages that weren't removed from the document**
image
### **the exact same copy but after loading a copy and deleting all of the deleted pages' objects from the document.**
image

**3-Other undamaged pdfs can be viewed better when copied to pdf-lib first:**

### **copy**
after6
### **original**
before6


mohamedsalem401:
@Hopding 
Hi there,
just wondering if this PR requires any modification to any documentation 🤔

src/api/PDFDocument.ts
```diff
@@ -147,6 +149,42 @@ export default class PDFDocument {
...
+        pdfCopy.setProducer(pdfDoc.getProducer()!);
+      }
+      if (pdfDoc.getSubject() !== undefined) {
+        pdfCopy.setSubject(pdfDoc.getSubject()!);
+      }
+      if (pdfDoc.getTitle() !== undefined) pdfCopy.setTitle(pdfDoc.getTitle()!);
+      pdfCopy.defaultWordBreaks = pdfDoc.defaultWordBreaks;
+
+      return pdfCopy;
+    }
```
Hopding, line 186: These changes are related to, but still distinct from, loading a document. So I think this would fit better as a `.copy()` method on `PDFDocument`. Then, instead of:
```js
PDFDocument.load(..., { returnCopy: true })
```
we could do:
```js
PDFDocument.load(...).copy()
```

Hopding: Once you've done this, be sure to add a proper doc comment to the `.copy()` method

Hopding (changes requested):
@mohamedsalem401 thanks for working on this! It's intriguing to me that building a copy seems to fix so many different issues. I wonder why that is? 🤔 

This is, of course, a workaround for the problems rather than a full solution of the root cause, because this requires users to know when they need to use `.copy()`. Also, `.copy()` won't always copy all information over to the new doc (acroforms, etc...).

Regardless, I'm happy to have a feature like this built into the lib. I've requested a slight refactor. Please also be sure to update the unit tests and integration tests to exercise this logic. 🙂

mohamedsalem401:
I have made the changes you asked for 😃 
I have also added documentation comment and tests for the new method.

Hopding (approved):
Looks great! Thanks again @mohamedsalem401! This feature will go out in the next release 🎉

mohamedsalem401:
Thanks for accepting my PR @Hopding 

I have some questions if you have the time:
1. I will try to add integration tests for the .copy() method. which pdf from the assets should I try to copy ? or should I create a new pdf ? 🤔
2. My PR should have closed #951 automatically when merged but since it was merged into a different branch the issue wasn't closed, should it be closed now or when the next release is here 🤔
3.  which issue in pdf-lib you would suggest I should try to work on next ?

Looking forward to making more contributions to this wonderful repository 😃

Hopding:
@mohamedsalem401 
1. I do request integration tests for most new features. But in this case, since the `copy` method packages up functionality that is exercised in other integration tests, I think just the unit tests will suffice.
2. Good point. I hadn't considered that. I think I'll drop a comment in that issue when the next release is cut, make sure it works for the OP, and close the issue if so.
3. Great question! I'm glad you're interested in making more contributions to pdf-lib! I don't have a good answer at the moment. But I'm working on a roadmap for pdf-lib and will be able to provide a better answer once it's done 🙂.

PR #901, opened 2021-06-02
Give access to ImageAlignment

oktapodia:
fix #900

PR #988, opened 2021-09-07
Give access to ImageAlignment

### LINKED ISSUE #900: Access to the ImageAlignment enum
Hello,

It looks like that we don't have access to the ImageAlignment class when importing the pdf-lib module.

##Current behavior

```
import {ImageAlignment } from 'pdf-lib'
        // ^ Cannot resolve symbol 'ImageAlignment'
```

##Expected behavior
```
import {ImageAlignment } from 'pdf-lib'
```

oktapodia:
fix #900

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -2,6 +2,7 @@ export * from 'src/api/form';
 export * from 'src/api/text';
 export * from 'src/api/colors';
 export * from 'src/api/errors';
+export * from "src/api/image";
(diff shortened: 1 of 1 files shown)


oktapodia:
Follow of https://github.com/Hopding/pdf-lib/pull/901 that should be mergeable @Hopding

Hopding (approved):
Looks solid! Thanks for taking care of this @oktapodia 👍

PR #991, opened 2021-09-14
Scale page content

JNK90:
I added two functions to the PDFPage to scale the content and the annots. This enables to resize the content and annots without violating private access modifiers. The code is inspired by the [Resizer Example](https://github.com/vegarringdal/simple-pdf-resizer).

### DIFF
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -40,6 +40,7 @@ import {
   PDFOperator,
   PDFPageLeaf,
   PDFRef,
+  PDFNumber
 } from 'src/core';
 import {
   addRandomSuffix,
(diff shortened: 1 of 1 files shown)


src/api/PDFPage.ts
```diff
@@ -566,6 +571,73 @@ export default class PDFPage {
...
+
+      ['RD', 'CL', 'Vertices', 'QuadPoints', 'L', 'Rect'].forEach((el) => {
+        const list = annot.get(PDFName.of(el)) as PDFArray;
+        if (list) this.scalePDFNumbers(list, x, y);
+      });
+
+      const pdfNameInkList = annot.get(PDFName.of('InkList')) as PDFArray;
+      pdfNameInkList
+        ?.asArray()
+        .forEach((arr) => this.scalePDFNumbers(arr as PDFArray, x, y));
```
Hopding, line 637: I need to improve the contributor docs on this, but:
* Please avoid using `asArray()` internally. It's just meant to be a convenience method for library consumers.
* Please also avoid using `.forEach()` and use a for-each-index loop instead. If you look through the codebase you'll see that's the pattern currently in use.

src/api/PDFPage.ts
```diff
@@ -566,6 +571,73 @@ export default class PDFPage {
...
+
+    const end = this.createContentStream(popGraphicsState());
+    const endRef = this.doc.context.register(end);
+
+    this.node.wrapContentStreams(startRef, endRef);
+  }
+
+  /**
+   * Scale the annots of a page. This is useful if you want to scale a page with comments or other annots.
+   * ```JS
```
Hopding, line 605:
```suggestion
   * ```js
```

src/api/PDFPage.ts
```diff
@@ -566,6 +571,73 @@ export default class PDFPage {
...
+   * // scale the content of the page down by 50% in x and y
+   * page.scaleContent(0.5, 0.5);
+   *
+   * // scale the content of the page down by 50% in x and y
+   * page.scaleAnnots(0.5, 0.5);
+   * ```
+   * See alos: [[scaleContent]]
```
Hopding, line 612:
```suggestion
   * See also: [[scaleContent]]
```

Hopding (changes requested):
This sounds neat! Thanks for working on it @JNK90 🙂 

I appreciate you adding good doc comments as well. That's something a lot of folks seem to forget about, haha.

In addition to my change request, can you:
* Explain a bit more about how this works? Please include screenshots/examples.
* Update one or two integration tests to showcase this behavior.

JNK90:
@Hopding I have added integration tests for node and web (I have problems to run the other environments on my system). I also replaced .forEach() and .asArray() calls.

To scale the content the method scaleContent(x, y) simply applys the scale-Operator to the contentstream. 
To scale the annots the method scaleAnnots(x, y) loops over all annots ('RD', 'CL', 'Vertices', 'QuadPoints', 'L', 'Rect', 'InkList') and multiplies the PDFNumbers of an annot with x or y as the factor.

I have edited a PDF and added, text, comments, shapes, highlights to test the scaling. 
With_annots
With_annots_scaled_content
With_annots_scaled_content_and_annots

Hopding (changes requested):
Great explanation, much appreciated @JNK90! Now that I have a good understanding of what this PR is for, I'm confident saying it'll be a heavily used feature!

I've requested a few changes below. Once these are addressed we should be good to go!

src/api/PDFPage.ts
```diff
@@ -566,6 +570,63 @@ export default class PDFPage {
...
+   * // bisect the size of the page
+   * p.setSize(p.getWidth() / 2, p.getHeight() / 2);
+   *
+   * // scale the content of the page down by 50% in x and y
+   * page.scaleContent(0.5, 0.5);
+   * ```
+   * @param x The factor by wich the x-axis for the content should be scaled (e.g. 0.5 is 50%)
+   * @param y The factor by wich the y-axis for the content should be scaled (e.g. 0.5 is 50%)
+   */
+  scaleContent(x: number, y: number): void {
```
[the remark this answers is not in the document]
Hopding: It would probably make sense to also call `setSize` within the high-level `scale` method. That way it becomes trivial for users to resize a page along with all of its contents.

src/api/PDFPage.ts
```diff
@@ -566,6 +570,63 @@ export default class PDFPage {
...
+    assertIs(x, 'x', ['number']);
```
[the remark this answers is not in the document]
JNK90: @Hopding this would limit user to only scale down a page. I think there is also a usecase for scaling a page up e.g. by 2 to double the size.

Hopding:
@JNK90 I see you marked all my comments as resolved, but I don't see any of the changes I requested in the diff? Did you forget to push your work? Or are you still working on this?

Hopding:
@JNK90 Ah, that would explain it!

PR #1002, opened 2021-09-23
Find and remove child widgets from the document context

### LINKED ISSUE #1001: Partial Appearance Left Behind After Field Removal
### What were you trying to do?

I was attempting to remove a field I had added previously using pdf-lib from a pdf and display that pdf with pdf.js in the browser.

I am using pdf-lib for the pdf alterations, and I am using pdf.js for displaying the pdf in-browser.

### Why were you trying to do this?

I am building a web tool that allows individuals to place / remove / adjust fields within a pdf. This is part of an automation system in a larger project.

### How did you attempt to do it?

I had previously added a field using pdf-lib to the pdf and saved it.

I then removed the field (can be textbox, checkbox, doesn't matter) and displayed the resulting pdf with pdf.js.

### What actually happened?

the default rectangle and border are still shown on the pdf. When I open the same pdf in Chrome pdf viewer or Acrobat, those appearances are not there. This leads me to suspect that something is being left behind when removing a field and that Chrome and Acrobat are intelligently ignoring whatever was still there. Pdf.js seems to still see this remnant and display it though.

### What did you expect to happen?

When the field is deleted and the pdf is reloaded into pdf.js, I would expect any visual trace of the field to be removed.

### How can we reproduce the issue?

I have attached a zip file with a single simple HTML document that creates a pdf document, adds some text and fields to said document, then removes a single field from that document before finally rendering that document to a canvas using pdf.js

[test.zip](https://github.com/Hopding/pdf-lib/files/7207547/test.zip)

### Version

1.16.0

### What environment are you running pdf-lib in?

Browser

### Required Reading

- [X] I have read [www.sscce.org](http://www.sscce.org/).
- [X] My report includes a Short, Self Contained, Correct (Compilable) Example.
- [X] I have read [Smart Questions](http://catb.org/~esr/faqs/smart-questions.html).
- [X] I have read [45 GitHub Issues Dos and Don'ts](https://davidwalsh.name/45-github-issues-dos-donts).

### Additional Notes

_No response_

Trapfether:
Fixes #1001 

When removing a field from the document, we need to remove the children of the field first; otherwise, some renderers will still see the remnants and display these dangling widgets.

## What?

Finds and removes child widgets of a field and removes them from the document context before removing field itself.

## Why?

Stops some renderers from still finding the dangling widgets and rendering their default appearances

## How?

In the removeField function, right before removing the field reference from the document context, we look for the field children by getting it's normalizedEntries. We check each child to see if it is indeed a reference; in which case, it is removed from the document context

## Testing?

I tested the changes against the automated testing suite and the integrated tests, especially integration test 18 which removes several fields as part of it's test.

## New Dependencies?

No

## Screenshots

[Before Fix.pdf](https://github.com/Hopding/pdf-lib/files/7214531/Before.Fix.pdf)
[After Fix.pdf](https://github.com/Hopding/pdf-lib/files/7214533/After.Fix.pdf)

## Suggested Reading?

Partly

## Anything Else?

Linter found style issue in the /src/api/index.ts file. I did not push that change as it was unrelated to this fix.

## Checklist

### DIFF
--- a/src/api/form/PDFForm.ts
+++ b/src/api/form/PDFForm.ts
@@ -596,6 +596,15 @@ export default class PDFForm {
 
     pages.forEach((page) => page.node.removeAnnot(field.ref));
     this.acroForm.removeField(field.acroField);
+    const docHandle = this.doc;
+    field.acroField
+      .normalizedEntries()
+      .Kids.asArray()
+      .forEach((child) => {
+        if (child instanceof PDFRef) {
+          docHandle.context.delete(child);
+        }
+      });
     this.doc.context.delete(field.ref);
   }
 


Hopding:
I saw you mentioned there was a failing integration test (over here https://github.com/Hopding/pdf-lib/issues/1001#issuecomment-925365062). Is that still the case, or were you able to fix it?

Trapfether:
@Hopding I was able to fix the issue. It was occurring because I was simply removing the appearance reference from the document context, but the widget reference itself was still registered and referring to the now missing reference.

This new approach looks for the widget reference itself by looping through the field's Kids and removes that instead.

This now passes all unit tests and I didn't spot any errors in the integration tests.

Hopding (changes requested):
@Trapfether this looks great! Thanks so much for making a PR. Lots of folks post issues, but it's relatively rare that anybody makes a PR to fix them 🙂 

My only ask is that you add some unit tests to exercise this logic. It would also be nice if you could update an integration test so that if there's a regression here we'll be more likely to catch it.

Trapfether:
I can absolutely do that.

A question on the Integration test. As the weird behavior only seemed to appear when using the pdf.js library to render the pdf, I would need to add that as a testing asset.

My question would be whether to make this another web integration test in the existing app:web suite, or should it be a new suite such as app:pdfjs.

As pdfjs is currently actively maintained, I don't think it hurts either way; however, the second method may be easier to deprecate should pdfjs stop being actively maintained at some point. There is also the question of which version of pdf.js to test against.

Alternatively, because we know that the visual aberration is the result of a dangling reference to a field's widget in the document context, I theoretically should be able to cover possible regressions through the unit test alone by testing whether child widget references are left after field removal.

Thank you for the encouragement and for taking the time to answer my questions.

Hopding:
@Trapfether Good question! Just adding it to the existing web suite should suffice. All of the web tests are run in Firefox (and [other browsers](https://github.com/Hopding/pdf-lib/blob/master/docs/RELEASING.md#checklist)) prior to cutting a release, and Firefox uses pdf.js to render PDFs. Of course, the test would no longer work if pdf.js is updated to correct the issue. But the unit tests will still prevent regressions within pdf-lib in that scenario. So just adding a bit of code to one existing integration test will do the trick. Note, however, that all the integration tests are kept in sync across all the apps. So you'll have to update the test in `apps/web`, `apps/node`, `apps/deno`, and `apps/rn` even if it's only verifiable in Firefox. This prevents the tests from fragmenting and becoming a maintenance burden.

P.S. I just noticed you're located in Wichita, KS. That's awesome! I grew up nearby in Derby, KS 🙂

Trapfether:
@Hopding I have added a unit test to PDFForm to specifically check for dangling child references.

Interestingly, we already had a test to do that in PDFForm; however, it loaded an existing PDF and removed the pre-existing fields from it. I essentially duplicated this test and created a PDF from scratch and added forms via the library instead.

I added a section in integration test 1 that adds and removes a field. There is text above the placement of this field saying "There should be no remnant of a field below this text". I made sure that text and field were the only thing on the bottom right of that page to make sure no future test additions should cause confusion.

Derby is awesome, I go to the theatre there every once in a while because they do throwback showings.

PR #1003, opened 2021-09-23
Set page reference for new field Widgets

Trapfether:
## What?

Sets the page reference for field widgets when they are added to a page

## Why?
Currently the library does not set the page reference, causing code intended to find the widget's page to fail. Notably, some internal code is also tripped up by this currently as it uses the same methodology.

Ex:
`

    const pages = PDFDoc.getPages();
    const field = PDFField;
    const widget = field.getWidgets()[0];
    const pageIndex = pages.findIndex(function(x){return x.ref === widget.P()});
    const page = pages[pageIndex];
`

## How?
A function has been added to PDFField **updateWidgetPageReference** that takes the widget annotation and the PDFRef for the page and adds that reference to the widget's dictionary.

This function is called at the end of a fields **addToPage** function just after adding the widget annotation to the page's internal node

## Testing?

I added a unit test for each field that supports an addToPage function. Upon making my changes, I verified that those tests now passed. I also checked the Integration tests to be sure nothing there was messed up. Everything appeared to be in order.

## New Dependencies?

No

## Screenshots

N/A

## Suggested Reading?

Partly

## Anything Else?

Linter found a lint rule violation in the src/api/index.ts file. I did not push that as it was unrelated to this fix

## Checklist
- [X] I added/updated unit tests for my changes.
- [X] I added/updated integration tests for my changes.

### DIFF
--- a/src/api/form/PDFButton.ts
+++ b/src/api/form/PDFButton.ts
@@ -181,6 +181,7 @@ export default class PDFButton extends PDFField {
 
     // Add widget to the given page
     page.node.addAnnot(widgetRef);
+    this.updateWidgetPageReference(widget, page.ref);
   }
 
   /**
--- a/tests/api/form/PDFCheckBox.spec.ts
+++ b/tests/api/form/PDFCheckBox.spec.ts
@@ -52,4 +52,20 @@ describe(`PDFCheckBox`, () => {
     expect(widgets().length).toBe(1);
     expect(widgets()[0].hasFlag(AnnotationFlags.Print)).toBe(true);
   });
+
+  it(`sets page reference when added to a page`, async () => {
+    const pdfDoc = await PDFDocument.create();
+    const page = pdfDoc.addPage();
+
+    const form = pdfDoc.getForm();
+
+    const checkBox = form.createCheckBox('a.new.check.box');
+
+    const widgets = () => checkBox.acroField.getWidgets();
+    expect(widgets().length).toBe(0);
+
+    checkBox.addToPage(page);
+    expect(widgets().length).toBe(1);
+    expect(widgets()[0].P()).toBe(page.ref);
+  });
 });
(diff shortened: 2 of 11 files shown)


src/api/form/PDFButton.ts
```diff
@@ -181,6 +181,7 @@ export default class PDFButton extends PDFField {
 
     // Add widget to the given page
     page.node.addAnnot(widgetRef);
+    this.updateWidgetPageReference(widget, page.ref);
```
Hopding, line 184: I definitely like the idea of setting the page reference for widgets. However, I'd prefer to add an optional `page` property to the `PDFField.createWidget` so it better fits with the existing API. So setting it would then look like:
```js
const widget = this.createWidget({
  width: options?.width ?? 100,
  height: options?.height ?? 50,
  ...,
  hidden: options?.hidden,
  page: page.ref,
});
```

tests/api/form/PDFCheckBox.spec.ts
```diff
@@ -52,4 +52,20 @@ describe(`PDFCheckBox`, () => {
     expect(widgets().length).toBe(1);
     expect(widgets()[0].hasFlag(AnnotationFlags.Print)).toBe(true);
   });
+
+  it(`sets page reference when added to a page`, async () => {
```
Hopding, line 56: Thanks for adding tests! 🎉

Hopding (changes requested):
Looks great overall! Just requested a small refactor.

src/api/form/PDFButton.ts
Trapfether: @Hopding I have reworked the interface to work as described.

As the function P() was defined on the PDFWidgetAnnotation Class, I added a setP function. This makes it so that PDFField is not directly accessing the widget dictionary, which should reduce coupling.

Now, in each context where createWidget is called, the page reference is passed as an option.

Is there someplace to document this low level API that I should be adding to?

PR #1032, opened 2021-10-14
Improved memory usage

Masterwow3:
We have been using pdf-lib for the last 4 months to generate our PDFs on the WEB and on the server. I noticed that for a PDF with 180 pages and high resolution PNGs, the RAM usage increases a lot (~6GB). 

My code change ensures that after a `page.drawImage` you can directly call an `image.embed()`.

This decodes and encodes the image immediately and remove the decoded image from RAM immediately.
With this change I have the same performance but a maximum RAM usage of 220MB.

Usage without these changes:
image
Max RAM usage:
image

Usage with these changes:
image
Max RAM usage:
image

### DIFF
--- a/src/api/PDFImage.ts
+++ b/src/api/PDFImage.ts
@@ -35,8 +35,8 @@ export default class PDFImage implements Embeddable {
   /** The height of this image in pixels. */
   readonly height: number;
 
-  private alreadyEmbedded = false;
-  private readonly embedder: ImageEmbedder;
+  private embedder: ImageEmbedder | undefined;
+  private embedderTask: Promise<PDFRef> | undefined;
 
   private constructor(ref: PDFRef, doc: PDFDocument, embedder: ImageEmbedder) {
     assertIs(ref, 'ref', [[PDFRef, 'PDFRef']]);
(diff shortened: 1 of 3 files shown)


Hopding (changes requested):
Hello @Masterwow3! Thanks for investigating this issue and opening a PR! I suspect this is a problem lots of people have run into, so it'll be great to release a fix 😃.

I like the general approach to took to eliminate the memory leak. However, I would like to change a few of the implementation details before merging:
* Please use `undefined` instead of `null`. The `null` value isn't used in the pdf-lib codebase.
* Instead of clearing the `PngEmbedder.image` property, we should clear the [`PDFImage.embedder`](https://github.com/Hopding/pdf-lib/blob/a936ad3de8df688b9f0830b490e246ee9e11c251/src/api/PDFImage.ts#L39) property after the image has been embedded.
* Please add unit tests to ensure that (1) the property is cleared after the first embed, and (2) a single `PDFDocument` object with an embedded image can still be updated and saved multiple times (unless there's already a test for this).

Let me know if you have any questions!

ritteratzuken:
Hello @Hopding,

i am pleased that the amendment is accepted. I have adapted the commit as you requested. 

The unneeded field "alreadyEmbedded" has been removed.
(https://github.com/Hopding/pdf-lib/pull/1032/files#diff-2e6b05e8490d2c6cc7ca78f78e8fa58bcaef509fbc97754da0808bca0f9d47ffL38)

Additionally, I made the embed() method parellel safe.
(https://github.com/Hopding/pdf-lib/pull/1032/files#diff-2e6b05e8490d2c6cc7ca78f78e8fa58bcaef509fbc97754da0808bca0f9d47ffR131)

Hopding (approved):
Looks great. Thanks @Masterwow3!

Issue #251, opened 2019-12-11
[Feature Request]: Option for `page.drawImage` to draw behind existing content on page


jtinbergen:
It is very useful when either merging PDF's or adding watermarks to PDFs to be able to draw an image on a page behind all the existing content instead of adding it over everything else at the end.

This is often useful when dealing with a PDF letter background you want to inject for company wide styling on documents.

mseshachalam:
@Hopding hey you done with this ? Can you point me how to do this ?

Hopding:
@mseshachalam I'm afraid I don't have time at the moment to create an example demonstrating how to do this. But essentially you'd need to change the position at which the content stream is inserted: 

https://github.com/Hopding/pdf-lib/blob/3c11ddfded4250e96c7e7f97c92c45754f9e7bee/src/api/PDFPage.ts#L1314

https://github.com/Hopding/pdf-lib/blob/3c11ddfded4250e96c7e7f97c92c45754f9e7bee/src/core/structures/PDFPageLeaf.ts#L119-L123

Rather than pushing the content stream to the end of the `Contents` array, you'd need to push it to the start, or perhaps somewhere in the middle.

uchamad:
Just want to share my solution to this scenario of adding background image to an existing page. It may be useful to someone looking for a solution.

```
async function addBackgroundImageToPdf(pdfBytes, imageUrl, outputFilePath) {
  const pdfDoc = await PDFDocument.load(pdfBytes);
  const {width, height} = pdfDoc.getPages()[0].getSize();

  const imageBytes = await fetch(imageUrl).then((res) => res.arrayBuffer());
  const backgroundImage = await pdfDoc.embedJpg(imageBytes);

  const copiedPages = await pdfDoc.copyPages(pdfDoc, [0])
  const [firstPage] = copiedPages;

  const embeddedPage = await pdfDoc.embedPage(firstPage);

  const newPage = pdfDoc.insertPage(0);
  newPage.drawImage(backgroundImage, {
    x: 0,
    y: 0,
    width: width,
    height: height,
    blendMode: 'Normal',
  });
  newPage.drawPage(embeddedPage);
  pdfDoc.removePage(1);

  const pdfBytesWithBackground = await pdfDoc.save();
  fs.writeFileSync(outputFilePath, pdfBytesWithBackground);
}
```

I am simply copying the contents of an existing page to a new page with an image background and then removing the original page.

yvesh:
[...]

PR #1059, opened 2021-10-28
Add place at back

Techno11:
## What?

Add ability to place drawn PDFPage elements in the back as opposed to the front

## Why?

I needed the ability to draw page elements behind elements that I've already drawn.

## How?

If the flag placeInBack is set on any drawable pdf element (like rectange, circle, etc) the drawn object is spliced into array position 0 of the content stream, rather than pushed to the end of the array.

## Testing?

I have tested drawing with all objects and ran the CI.

## New Dependencies?

N/A

## Screenshots

image
The red rectangle was drawn after the text that is on top of it

## Suggested Reading?

Yes

## Anything Else?

N/A

## Checklist
- [X] I added/updated doc comments for any new/modified public APIs.

### DIFF
--- a/apps/deno/tests/test1.ts
+++ b/apps/deno/tests/test1.ts
@@ -646,6 +646,63 @@ export default async (assets: Assets) => {
 
   form.removeField(textField);
 
+  /********************** Page 6 **********************/
+
+  const page6 = pdfDoc.addPage([size, size]);
+
+  page6.drawText('This was drawn before the red square, but should be on top of it!', {
+    x: 0,
+    y: 120,
+    font: ubuntuFont,
+    size: 14
+  });
+  page6.drawSquare({
+    size: size / 4,
+    color: rgb(1, 0, 0),
+    y: 100,
+    placeBehindAll: true
+  });
+
+  // Draw a stack of objects in reverse
+  page6.drawText('These objects were drawn from smallest to largest, using placeBehindAll', {
+    y: (size/2) + 260,
+    x: (size/2) - 150,
+    font: ubuntuFont,
+    size: 14
+  });
+  page6.drawCircle({
+    size: size / 10,
+    color: rgb(1, 0, 0),
+    y: size/2,
+    x: size/2,
+    placeBehindAll: true
+  });
+  page6.drawEllipse({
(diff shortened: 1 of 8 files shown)


Hopding (changes requested):
Hello @Techno11! This is a cool idea! Thanks for making a PR 😃.

A couple things I'd like to change before merging:
* The term `placeAtBack` isn't ideal for search indexing, and I suspect many users will find it unintuitive. Naming is hard, but I think `underneath` will be more indexable and intuitive. (`behind` and `below` are two other alternatives that come to mind).
* Please update an existing integration test to showcase this behavior for 2 or three of the `PDFPage` methods you've updated.
* Please update the [`PDFContentStream.spec.ts`](https://github.com/Hopding/pdf-lib/blob/master/tests/core/structures/PDFContentStream.spec.ts) file to test the new `.splice(...)` method.

Techno11:
Hi @Hopding, thanks for the quick review. I added the .splice() to the spec file, as well as added a 6th page to test1, which looks like the following:
image


Here is the full PDF of the modified test1
[pdf_with_drawBehind_objs.pdf](https://github.com/Hopding/pdf-lib/files/7453906/pdf_with_drawBehind_objs.pdf)

Lastly, (I forgot to include this in the original comment) this implements and closes #251 
(even though it is stale and has already been closed)

Hopding:
@Techno11 thanks for making the changes! Can you please explain how this functionality works when used on an existing PDF file and update a second integration test to demonstrate this?

chebum:
@Hopding I believe the suggested approach won't work when modifying existing document: PDFPage.getContentStream() inserts the newly created content stream on top of the page existing content: 

    this.node.addContentStream(this.contentStreamRef);

Should one need to insert an image before existing content, this code can be used:

    page.drawImage(imageRef, { /* x, y, width, height */ });
    const contents = page.node.normalizedEntries().Contents;
    const newContentsRef = contents.get(contents.size() - 1);
    contents.remove(contents.size() - 1);
    contents.insert(0, newContentsRef);

It works with current version of pdf-lib. No modifications necessary.

Techno11:
Closing as the above solution works. Thanks!

PR #1216, opened 2022-04-17
Provide all text state operators

retfah:
## What?
The pdf specification allows several text state operators, e.g. for the font, size (Tf) and the line distance (leading, TL). However, there are some more (text rise, horiztontal scaling, word spacing, character spacing, text rendering mode) and I want to make use especially of the horizontalScaling (Tz). I saw that in the OperatorNames actually all were already defined, but not provided to the user (e.g. in drawText) yet. Therefore, I extended everything needed so that all text state operators can be set in the options for drawText. All those operators are implemented to be written only if the option is actually set, i.e. no default values are written to pdf, which would make the file unnecessarily larger.
Example:
`page.drawText('hello world', {
    y: size - 40,
    size: 20,
    lineHeight: 20,
    horizontalScale: 50,
    rise: 5,
    wordSpace: 10,
    renderingMode: TextRenderingMode.Outline,
    characterSpace: 4,
  });`

## Why?
I will need horizontal text scaling. Other users might want to use this or any of the other 4 added text state operators (word spacing, character spacing, text rise, text rendering mode) as well.

## How?
It is implemented in the exact sam way as setting the font (Tf) or the lineHeight (TL). It is the most logic implementation and follows the present code structure. The horizontal spacing was actually already implemented not just in OperatorNames, but also the operator was already written in the name of "setCharacterSqueeze". In my opinion, it would be nicer if the operator would use the name as in the standard (i.e. setHorizontalScaling), as it is the case for all the other operators I have looked at. Therefore, created an additional operator "setHorizontalScaling", but kept the old "setCharacterSqueeze" one for backwards compatibility (with a deprecated note in the docstring). (Note: setCharacterSqueeze was used neither in the code nor in the tests (at least not uncommented), but users might by using it "the hard way", bypassing the "drawText"-function. 

## Testing?
Yes. 
Unit tests: All pass. 
Integration tests: added a page in test 1 with the new options. All pass. pdf looks ok in Adobe, Foxit, Chrome, Firefox.
Linting successful. 
Typecheck successful.

The changes to the code can hardly have an impact to any existing code, since it only extends the existing part. The added options are working as they should, which can be seen in the last page of test1.

## New Dependencies?
No.

## Screenshots
N/A

## Suggested Reading?
Yes

## Anything Else?
I did not add any unit tests, since there were no unit tests for the other text state operators and I haven't created any unit tests so far (being just a hobby coder). For this reason, I could not just create an analogous test for the added text state options. 

## Checklist
- [X] I added/updated integration tests for my changes.
- [X] I added/updated doc comments for any new/modified public APIs.

### DIFF
--- a/apps/deno/tests/test1.ts
+++ b/apps/deno/tests/test1.ts
@@ -646,6 +646,67 @@ export default async (assets: Assets) => {
 
   form.removeField(textField);
 
+  /********************** Page 6 **********************/
+  // This page tests different drawing operations as well as adding custom
+  // operators to the page content.
+
+  const page6 = pdfDoc.addPage([size, size]);
+
+  const text = "These are the test words. "
+  page6.drawText(text + 'regular', {
+    y: size - 20,
+    size: 20,
+    lineHeight: 20,
+    //horizontalScale: 50,
+    //rise: 0,
+    //wordSpace: 10,
+    //characterSpace: -2,
+  });
+  page6.drawText(text + '+5 raised and fontsize 12 instead of 20', {
+    y: size - 20,
+    x: 325,
+    size: 12,
+    lineHeight: 20,
+    //horizontalScale: 50,
+    rise: 5,
+    //wordSpace: 10,
+    //characterSpace: -2,
+  });
+  page6.drawText(text + '50% horizontal scale', {
+    y: size - 40,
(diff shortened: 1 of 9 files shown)


scolastico:
would love to see that merged

Issue #1267, opened 2022-07-08
Flattened PDFs Malformed


kyle-leitham-aix:
### What were you trying to do?

Print a flattened PDF in Adobe.

### How did you attempt to do it?

Download the PDF generated in the flatten code example provided in the README: https://jsfiddle.net/Hopding/skevywdz/2/ and attempt to print it in an Adobe PDF reader.

### What actually happened?

The PDF won't print and Adobe will give you error dialogs:
<img width="515" alt="Screen Shot 2022-07-08 at 10 32 57 AM" src="https://user-images.githubusercontent.com/106110469/178013361-e2661072-dd3d-45d2-b6af-fc3484b17a08.png">
<img width="388" alt="Screen Shot 2022-07-08 at 10 33 03 AM" src="https://user-images.githubusercontent.com/106110469/178013366-1fcc915f-ad30-4956-bb39-5eec8f35fc8d.png">
:

### What did you expect to happen?

The PDF should be able to print

### How can we reproduce the issue?

- Download the PDF generated in the flatten code example provided in the README: https://jsfiddle.net/Hopding/skevywdz/2/

- Attempt to print it in an Adobe Acrobat Reader. Or, if you have an Adobe Acrobat Pro license, you can open the PDF and go to Preflight > PDF Analysis > Report PDF Syntax Issues and click Analyze. You should see these syntax errors describing that the page annotations are pointing to null references:
<img width="813" alt="Screen Shot 2022-07-08 at 10 38 54 AM" src="https://user-images.githubusercontent.com/106110469/178014365-624a5f12-0d92-497d-9606-961b8a87873d.png">

### Version

v1.17.1

### What environment are you running pdf-lib in?

Node

### Checklist

- [X] My report includes a Short, Self Contained, Correct (Compilable) Example.
- [X] I have attached all PDFs, images, and other files needed to run my SSCCE.

PR #1268, opened 2022-07-08
Getting correct widget ref to be removed

kyle-leitham-aix:
## What?

Gets correct widget ref from the document to remove the corresponding annotation ref from the page
## Why?

Trying to fix this issue: https://github.com/Hopding/pdf-lib/issues/1267
## How?

Gets the ref using doc.context.getObjectRef instead of findWidgetAppearanceRef (which seems to get the wrong ref?)
## Testing?

Ran the unit tests

## New Dependencies?

N/A
## Screenshots

N/A
## Suggested Reading?

No
## Anything Else?

I'm new to this repo and PDF specifications, so I'm not sure this is the correct fix. Hoping it is!
## Checklist
- [ ] I [ran the linter](https://github.com/Hopding/pdf-lib/blob/master/docs/CONTRIBUTING.md#running-the-linter) on my changes.

### DIFF
--- a/src/api/form/PDFForm.ts
+++ b/src/api/form/PDFForm.ts
@@ -584,12 +584,14 @@ export default class PDFForm {
 
     for (let i = 0, len = widgets.length; i < len; i++) {
       const widget = widgets[i];
-      const widgetRef = this.findWidgetAppearanceRef(field, widget);
+      const widgetRef = this.doc.context.getObjectRef(widget.dict);
 
       const page = this.findWidgetPage(widget);
       pages.add(page);
 
-      page.node.removeAnnot(widgetRef);
+      if (widgetRef !== undefined) {
+        page.node.removeAnnot(widgetRef);
+      }
     }
 
     pages.forEach((page) => page.node.removeAnnot(field.ref));


mel-habip:
@kyle-leitham-aix do you know if there is any progress on this? I'm also running into this issue and marking the fields read-only isn't really an option since they're still editable. Any further info?

kyle-leitham-aix:
@mel-habip It seems like this repo has been abandoned, unfortunately. As a workaround, I had to programmatically remove all of the page annotations before flattening.

mpalmahsf:
> @mel-habip It seems like this repo has been abandoned, unfortunately. As a workaround, I had to programmatically remove all of the page annotations before flattening.

Could you provide an example of how to do this?  We tried this but didn't quite work out for us.

kyle-leitham-aix:
@mpalmahsf So, if you have the pdf document instance, it would be something like:
```
pdfDocument.getPages().forEach((page) => {
      const annotations = page.node.Annots()
      const size = annotations?.size()
      if (size) {
        for (let i = 0; i < size; i++) {
          annotations.remove(annotations.get(i))
        }
      }
    })
```

satyajitnayk:
There is an issue with this approach. 
After this fix adobe no more throw any error but all fields & texts are missing from the pdf(Just an empty PDF page).

dasaevc:
> @mpalmahsf So, if you have the pdf document instance, it would be something like:
> 
> ```
> pdfDocument.getPages().forEach((page) => {
>       const annotations = page.node.Annots()
>       const size = annotations?.size()
>       if (size) {
>         for (let i = 0; i < size; i++) {
>           annotations.remove(annotations.get(i))
>         }
>       }
>     })
> ```

I tried this approach and worked but I applied that AFTER flattening the form. In that way the fields and texts appear well.

PR #1379, opened 2023-01-14
feat: add page translate function

mo-bai:
## What?
I added one [translate] function to the PDFPage class . the method can translate annotations and content of the PDF's page . not only then content .

```js
let source = await PDFDocument.load(readFile(path))
const pdfDoc = await PDFDocument.create()
let pages = await pdfDoc.copyPages(source, source.getPageIndices())
for (let page of pages) {
page.translate(50,50)
pdfDoc.addPage(page)
const pdfBytes = await pdfDoc.save()
```

## Why?
In many pdf files, annotations and text content is followed together, the library currently only provides the translateContent method, if you use the method will cause misalignment of annotations and text . also , there are no methods in the library for moving annotations

## How?
In my method, I first find all the annotations and then trnaslate the annotations by  adding or subtracting numbers from the location information in the annotations

## Testing?
I created a pdf file, marked the annotation on top of a word, and then use my method  [ translate(20,20) ] , and then check if the annotation deviates from the word . Result: No offset . then repeat doing this test by using different annotation type

## New Dependencies?
no

## Screenshots

the yellow line is annotation

## Suggested Reading?
Yes , but the url(https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf) is 404

## Checklist
- [x] I added/updated doc comments for any new/modified public APIs.

### DIFF
--- a/src/api/PDFPage.ts
+++ b/src/api/PDFPage.ts
@@ -569,6 +569,31 @@ export default class PDFPage {
   }
 
   /**
+   * translate this page's content and annotations. In some scenarios, the annotations
+   * is following this text.If only the [translateContent] function is used, 
+   * the position of the annotations relative to the text will change,so you can use [translate] function .
+   * For example:
+   * ```js
+   * // Move the page's content and annotations from the lower-left corner of the page
+   * // to the top-right corner.
+   * page.translate(50, 50)
+   * ```
+   * @param x The new position on the x-axis for this page's content.
+   * @param y The new position on the y-axis for this page's content.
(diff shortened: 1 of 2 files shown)


anodynos:
@mo-bai please check my comment [here](https://github.com/Hopding/pdf-lib/issues/1423#issuecomment-1492631262)

Sharcoux (changes requested):
Missing:
- tests
- documentation

Issue #1349, opened 2022-11-18
Error: Could not find page for PDFRef 270 0 R


enoh-barbu:
### What were you trying to do?

Flatten pdf form

### How did you attempt to do it?

const form = pdfDoc.getForm();
form.flatten();

### What actually happened?

I received an error:

```
Error: Could not find page for PDFRef 270 0 R
    at PDFForm.findWidgetPage (/usr/app/node_modules/pdf-lib/src/api/form/PDFForm.ts:713:15)
    at PDFForm.flatten (/usr/app/node_modules/pdf-lib/src/api/form/PDFForm.ts:550:27)
    at /usr/app/src/index.ts:116:18
    at Generator.next (<anonymous>)
    at fulfilled (/usr/app/src/index.ts:5:58)
```

### What did you expect to happen?

I should not receive that error

### How can we reproduce the issue?

I can't provide my pdf document, but it seems that the pdf is containing some strange/bkoken form fields

### Version

1.17.1

### What environment are you running pdf-lib in?

Node

### Checklist

- [X] My report includes a Short, Self Contained, Correct (Compilable) Example.
- [X] I have attached all PDFs, images, and other files needed to run my SSCCE.

bruno-panuto-embroker:
seeing this too.

Issue #1281, opened 2022-07-20
Could not find page for PDFRef Error during flatten


makarov-roman:
### What were you trying to do?

flatten document.

### How did you attempt to do it?

example to reproduce below
### What actually happened?

crash on flatten call
```
Error: Could not find page for PDFRef 220 0 R
    at PDFForm.findWidgetPage (/home/roman/Projects/monorepo/packages/lambdas/pdfService/node_modules/pdf-lib/cjs/api/form/PDFForm.js:603:23)
    at PDFForm.flatten (/home/roman/Projects/monorepo/packages/lambdas/pdfService/node_modules/pdf-lib/cjs/api/form/PDFForm.js:458:33)
```

### What did you expect to happen?

not having a crash

### How can we reproduce the issue?

```js
const { PDFCheckBox, PDFDocument, PDFTextField } = require('pdf-lib');
const fs = require('fs');

const fill = async () => {
  try {
    const Body = fs.readFileSync(
      'internshipApplicationForFurtherEducation.pdf',
    );

    const pdfDoc = await PDFDocument.load(Body);
    const form = pdfDoc.getForm();
    form.flatten();

    const result = await pdfDoc.save({ useObjectStreams: true });
    fs.writeFileSync('result.pdf', result);
  } catch (e) {
    console.error(e);
    throw e;
  }
};
fill();

```
[file](https://drive.google.com/file/d/1uhmiwcyESdbmuIPRLs3i6kQSGNtIJDCS/view?usp=sharing)

### Version

1.17.1

### What environment are you running pdf-lib in?

Node

### Checklist

- [X] My report includes a Short, Self Contained, Correct (Compilable) Example.
- [X] I have attached all PDFs, images, and other files needed to run my SSCCE.

### Additional Notes

In our project I have around 30 different pdfs and only 2 has such issue. Not sure if it's corrupted metadata or something else. Before it was working with pdftk.

makarov-roman:
I've made minor patch to ignore this error inside `flatten` and `findWidgetPage` functions and result doc look fine. So my assumption is that some editor left broken meta-data, thus pdflib "see" widget, but cannot find it on page. Ideally we should check it during `load` and clean such broken widgets/refs.
I'll publish the workaround tomorrow.

makarov-roman:
[here is my workaround](https://github.com/Hopding/pdf-lib/commit/f7965b9f1aef688f517d38252ba6328e76b74783)
[...]

PR #1407, opened 2023-02-26
Make pdf-lib more graceful like other pdf software

### LINKED ISSUE #967: Error when using flatten, Could not find page for PDFRef
I'm currently trying to figure out what is going on.

I fill a form from my PDF and then I need to use flatten in order to "save" the values and then download said PDF.

At the moment of using flatten there is an Unhandled Rejection, as follows

`Unhandled Rejection (Error): Could not find page for PDFRef 439 0 R`

which I understand means that once the field is flatten, the library will delete the field and uses the ref value of the field in order to do this, in this case, it looks for a field with a ref value of "439 0 R" but it does not exist, so I looked around to see what was said problem, using the following code

```
const fields = form.getFields()
fields.forEach(field => {
    const type = field.constructor.name
    const name = field.getName()
    const ref = field.ref
    const acroField = field.acroField
    console.log(`${type}: ${name} - ${ref}`)
  })
```
and found out that the ref values only go around 200-ish, about the same number of fields from the PDF.

```
PDFCheckBox: cb_1_1_a - 11 0 R 
PDFCheckBox: cb_1_1_b - 12 0 R
PDFCheckBox: cb_1_1_c - 13 0 R 
PDFCheckBox: cb_1_1_d - 14 0 R
PDFCheckBox: cb_1_1_e - 15 0 R 
PDFCheckBox: cb_1_1_f - 16 0 R 
PDFCheckBox: cb_1_1_g - 17 0 R 

[...]

PDFCheckBox: cb_52_4_s - 208 0 R
PDFCheckBox: cb_52_4_n - 209 0 R 
PDFTextField: tb_53_1 - 210 0 R 
PDFTextField: tb_53_2 - 211 0 R 
PDFTextField: tb_53_3 - 212 0 R
PDFTextField: tb_53_4 - 213 0 R
PDFTextField: tb_53_5 - 214 0 R 
PDFCheckBox: cb_1_8_b - 215 0 R
PDFCheckBox: cb_1_8_c - 216 0 R
```

the field "cb_1_8_c " is the last field of the form and has a ref value of "216 0 R"

So, I have no idea what is happening, what is causing the function flatten to use a ref value of "439 0 R" when none of the current fields have such.

The PDF I'm using was a flat document that I later add field forms using online tools.

emilsedgh:
Some broken fields may include annotations in their Kids array that couldn't be found on any pages. Up until now we've been throwing an exception when dealing with such fields. But it appears that other PDF software are more resilient to this and gracefully ignore them.

This commit ensures we'll do the same.

Fixes #967,#1281,#1349

## What?

PDF Fields can have `Kids` in them, each `Kid` being a `WidgetAnnotation`. In some rare cases (probably due to bad pdf generators) there are cases that there is a `PDFField` with a an annotation in it `Kids` but that annotation doesn't exist on any page. Therefore we don't know on which page it must be rendered.

## Why?

`PDFForm.findPageForAnnotationRe` is responsible for finding the page for a given annotation. Up until now it'd throw an exception if it couldn't find a page.

This means when trying to flatten such PDF's, pdf-lib would throw an error. So basically you have a pdf file that can be opened up with Chrome, Firefox/pdfjs, Acrobat Reader and even pdf-lib. But trying to flatten it using pdf-lib would cause an exception (as seen in #967,#1281,#1349)

## How?

This patch makes pdf-lib's Flatten gracefully ignore such cases and render the PDF like other pdf readers.

## Checklist
- [x] I added/updated doc comments for any new/modified public APIs.

### DIFF
--- a/src/api/form/PDFForm.ts
+++ b/src/api/form/PDFForm.ts
@@ -705,7 +705,7 @@ export default class PDFForm {
     return this.defaultFontCache.access();
   }
 
-  private findWidgetPage(widget: PDFWidgetAnnotation): PDFPage {
+  private findWidgetPage(widget: PDFWidgetAnnotation): PDFPage | undefined {
     const pageRef = widget.P();
     let page = this.doc.getPages().find((x) => x.ref === pageRef);
     if (page === undefined) {
--- a/src/api/form/PDFForm.ts
+++ b/src/api/form/PDFForm.ts
@@ -548,6 +548,9 @@ export default class PDFForm {
       for (let j = 0, lenWidgets = widgets.length; j < lenWidgets; j++) {
         const widget = widgets[j];
         const page = this.findWidgetPage(widget);
+        if (page === undefined) {
+          continue;
+        }
         const widgetRef = this.findWidgetAppearanceRef(field, widget);
 
         const xObjectKey = page.node.newXObject('FlatWidget', widgetRef);
@@ -587,6 +590,10 @@ export default class PDFForm {
       const widgetRef = this.findWidgetAppearanceRef(field, widget);
 
       const page = this.findWidgetPage(widget);
+      if (page === undefined) {
+        continue;
+      }
+
       pages.add(page);
 
       page.node.removeAnnot(widgetRef);
@@ -709,8 +716,13 @@ export default class PDFForm {
 
       page = this.doc.findPageForAnnotationRef(widgetRef);
 
+      /*
+       * In some broken PDF's there may be widgets that don't exist on any page.
+       * We used to throw an error in such cases but it appears that other
+       * PDF readers just ignore such PDF's.
+       */
       if (page === undefined) {
-        throw new Error(`Could not find page for PDFRef ${widgetRef}`);
+        return undefined;
       }
     }
 


anodynos:
@emilsedgh please check my comment [here](https://github.com/Hopding/pdf-lib/issues/1423#issuecomment-1492631262)

satyajitnayk:
Hey @Hopding. It would be great if you could review this PR. It is an important fix & requested by many.

caub:
cf https://github.com/Hopding/pdf-lib/issues/1281 too

jalbertsr:
Is this planning to be merged? I'm currently using this alternative fork [@visaright/pdf-lib](https://www.npmjs.com/package/@visaright/rjsf-utils) only because of this issue.

theprogramsam:
Can we get this in please? @Hopding

aktary:
Bump. @Hopding

sstepanchuk:
Bump. @Hopding

iway1:
@anodynos what does your comment have to do with this? @cantoo/pdf-lib also has this problem

sstepanchuk:
Here is my workaround, maybe it will be helpful

```typescript
 const form = pdfDoc.getForm();
    if (form && form.getFields().length > 0) {
      const pages = pdfDoc.getPages();
      const formFields = form.getFields();
      const firstPageRef = pages[0].ref;

      for (const field of formFields) {
        const widgets = field.acroField.getWidgets();
        for (const widget of widgets) {
          const widgetPageRef = widget.P();

          // If widget has no page reference, assign it to first page
          if (!widgetPageRef) {
            try {
              widget.setP(firstPageRef);
            } catch {}
          }
        }
      }
      form.flatten();
    }
```

Waxo:
@Hopding this is needed

PR #1500, opened 2023-08-06
Updated yarn add in README.md

s0h311:
## What?
updated the installation command for yarn.

## Why?
For dev dependency it should be `yarn add -D` and not just `yarn add`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1127,7 +1127,7 @@ To install the latest stable version:
 npm install --save pdf-lib
 
 # With yarn
-yarn add pdf-lib
+yarn add -D pdf-lib
 ```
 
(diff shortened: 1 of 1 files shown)


ghost (approved):

thm76:
I think this needs to be more nuanced because in some cases you will need to install pdf-lib as a dependency (I'm generating PDFs inside a React App) but in other cases it's a `devDependency` - I have another app where I'm generating PDFs at build time only.

In my opinion I would leave it as is, no "-D" flag. This is also consistent with the `npm install` command.

If the "-D" flag should be added then add it to `npm install` as well.

PR #1653, opened 2024-07-17
Update PDFObjectParser.ts

UnderKoen:
## What?
Closes an memory leak.

## Why?
When combing many large PDF's the memory usage is through the roof.

## How?
By creating an new buffer instead of an view the reference to the orginal buffer is removed and can be picked up on garbage collection.

## New Dependencies?
No

## Screenshots
N/A

## Suggested Reading?
No

## Checklist
	- I could not build the project locally? No clue why. Same issue as https://github.com/Hopding/pdf-lib/discussions/1630
- [z] My changes work for both **new** and **existing** PDF files.

### DIFF
--- a/src/core/parser/PDFObjectParser.ts
+++ b/src/core/parser/PDFObjectParser.ts
@@ -236,7 +236,7 @@ class PDFObjectParser extends BaseParser {
       end = this.findEndOfStreamFallback(startPos);
     }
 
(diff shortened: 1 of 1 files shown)


UnderKoen:
After some more testing I don't think this is the solution to the problem. When an Buffer (Node) is supplied this is an issue because Node `slice` returns an view.

PR #1722, opened 2025-01-23
Fix NaN validation

SkyaTura:
## What?
This PR fixes a bug on utility validators where any parameters validated with `assertIs` would fail asserting Arrays because of unguarded `isNaN` call.

## Why?
I don't really know how this is working at all for anyone, since I don't know how `isNaN` could return false for anything but a number.

## How?
1. Don't use `globalThis.isNaN` since it is meant to check values of ANY type. Use `Number.isNaN` instead. This alone should do the trick anyway.
2. Move the isNaN clause inside the typeof number clause. This is solely for optimization...

## Testing?
I'm pretty confidente that the this PR won't break anything. What I don't know is how is working anywhere, haha

## Suggested Reading?

https://stackoverflow.com/questions/33164725/confusion-between-isnan-and-number-isnan-in-javascript

## Checklist
- [ ] I [ran the linter](https://github.com/Hopding/pdf-lib/blob/master/docs/CONTRIBUTING.md#running-the-linter) on my changes.

### DIFF
--- a/src/utils/validators.ts
+++ b/src/utils/validators.ts
@@ -75,8 +75,10 @@ export const getType = (val: any) => {
   if (val === null) return 'null';
   if (val === undefined) return 'undefined';
   if (typeof val === 'string') return 'string';
-  if (isNaN(val)) return 'NaN';
-  if (typeof val === 'number') return 'number';
+  if (typeof val === 'number') {
+    if (Number.isNaN(val)) return 'NaN';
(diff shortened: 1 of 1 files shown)


Sharcoux:
Will be solved in v2.5.3 of [@cantoo/pdf-lib](https://github.com/cantoo-scribe/pdf-lib/pulls)

PR #1755, opened 2025-08-10
fix issue #951 - Introduced logic for output reordering of indirect o…

parvit:
## What?
This PR fixes issue #951.

## How?
Issue pops up in files which use the "Incremental Update" feature and have modifed indirect objects in the updated sections but no new objects added.
If the operation of sorting by object number is done for this kind of documents in the created context, than the objects are written at the wrong offsets and readers become confused leading to the rendering errors.
In case instead where new objects are added to the file, the sorting is necessary to consolidate the changes. At least that is what i've observed as simply removing the sorting lead to tests 17 and 18 to become corrupted (because those tests load a file with multiple update sections and add new objects).

## Testing?
Tested by introducing a new test19 integration test which loads the problematic file and then executes the save operation to show the produced pdf file to the browser.

## New Dependencies?
No

## Screenshots
N/A

## Checklist
- [x] I tested my changes in the browser.


parvit:
@Hopding could you review the change?

Sharcoux:
Released in [@cantoo/pdf-lib](https://www.npmjs.com/package/@cantoo/pdf-lib) (At least I believe so).

PR #1780, opened 2026-04-12
fix: add decompression size limit to prevent decompression bomb DoS

### LINKED ISSUE #1777: Unbounded buffer growth in DecodeStream enables decompression bomb DoS
## Summary

`DecodeStream.ensureBuffer()` in `src/core/streams/DecodeStream.ts` (lines 148-160) doubles its internal buffer without any upper bound. A crafted PDF containing a compressed stream with a high compression ratio (decompression bomb) can force `PDFDocument.load()` to allocate gigabytes of memory, causing an out-of-memory crash.

## Vulnerability Details

**CWE-409: Improper Handling of Highly Compressed Data (Decompression Bomb)**

The `ensureBuffer()` method doubles the buffer size in a loop until it meets the requested size, with no maximum:

```typescript
protected ensureBuffer(requested: number) {
  const buffer = this.buffer;
  if (requested <= buffer.byteLength) {
    return buffer;
  }
  let size = this.minBufferLength;
  while (size < requested) {
    size *= 2;
  }
  const buffer2 = new Uint8Array(size);
  buffer2.set(buffer);
  return (this.buffer = buffer2);
}
```

This is the base class for all decompression streams (FlateStream, LZWStream, etc.). The `decode()` method (line 139-141) and `getBytes()` without a length argument (lines 93-96) both call `readBlock()` in a loop until EOF, with each block potentially doubling the buffer:

```typescript
decode(): Uint8Array {
  while (!this.eof) this.readBlock();
  return this.buffer.subarray(0, this.bufferLength);
}
```

### Attack scenario

1. Attacker crafts a PDF with a Flate-compressed object stream (ObjStm) or XRef stream containing data that achieves a very high compression ratio (e.g., repeated zero bytes compress extremely well with deflate — a 10KB compressed payload can decompress to several GB).
2. When `PDFDocument.load()` parses the PDF, it decompresses the stream to read object definitions.
3. `ensureBuffer()` allocates exponentially growing buffers: 512 → 1KB → 2KB → ... → 1GB → 2GB → 4GB.
4. The Node.js process runs out of memory and crashes, or in browser contexts, the tab is killed.

This is triggered automatically during normal PDF loading — no special API usage required.

### Proof of concept

```javascript
const { PDFDocument } = require('pdf-lib');
const fs = require('fs');
const zlib = require('zlib');

// Create a minimal PDF with a decompression bomb in an object stream
// The deflated payload is ~10KB but decompresses to ~1GB
const zeros = Buffer.alloc(100 * 1024 * 1024, 0); // 100MB of zeros
const compressed = zlib.deflateSync(zeros);
console.log(`Compressed: ${compressed.length} bytes -> decompresses to 100MB`);

// In a real attack, this compressed data would be embedded as a 
// FlateDecode stream in the PDF's object stream or XRef stream.
// PDFDocument.load() would decompress it without any size check.
```

### Impact

- **Denial of service**: Any application that accepts user-uploaded PDFs and processes them with `PDFDocument.load()` is vulnerable. A small malicious PDF (< 100KB) can consume all available memory.
- **Server-side risk**: Node.js servers using pdf-lib to merge, modify, or validate uploaded PDFs can be crashed by a single request.

### Suggested fix

Add a maximum buffer size limit to `ensureBuffer()`:

```typescript
private static readonly MAX_BUFFER_SIZE = 256 * 1024 * 1024; // 256 MB

protected ensureBuffer(requested: number) {
  const buffer = this.buffer;
  if (requested <= buffer.byteLength) {
    return buffer;
  }
  if (requested > DecodeStream.MAX_BUFFER_SIZE) {
    throw new Error(
      `DecodeStream buffer size limit exceeded: requested ${requested} bytes ` +
      `(max ${DecodeStream.MAX_BUFFER_SIZE}). The stream may be a decompression bomb.`
    );
  }
  let size = this.minBufferLength;
  while (size < requested) {
    size *= 2;
  }
  const buffer2 = new Uint8Array(size);
  buffer2.set(buffer);
  return (this.buffer = buffer2);
}
```

Ideally, this limit should be configurable via `PDFDocument.load()` options so users can adjust it for their use case.

## Version

Tested against pdf-lib 1.17.1 (latest).

## Environment

Node.js

eddieran:
## Summary

Adds a configurable maximum decompressed size limit to `DecodeStream` to prevent decompression bomb denial-of-service attacks.

**Problem:** A crafted PDF can contain a small compressed stream that expands into gigabytes of data when decompressed, exhausting the host's available memory (decompression bomb / zip bomb). The `DecodeStream` base class had no upper bound on buffer growth — `ensureBuffer()` would keep doubling the internal buffer indefinitely.

**Fix:** Enforce a maximum decompressed size (default 100 MB) in `ensureBuffer()`. When the accumulated output exceeds this threshold, an error is thrown instead of continuing to allocate memory. The limit is configurable via the `maxDecompressedSize` constructor parameter for users who need to process legitimately large streams.

### Changes

- Added `DEFAULT_MAX_DECOMPRESSED_SIZE` constant (100 MB)
- Added `maxDecompressedSize` parameter to `DecodeStream` constructor (with default)
- Added size check in `ensureBuffer()` that throws before allocating beyond the limit

### Backward Compatibility

The new parameter has a default value, so all existing subclasses (`FlateStream`, `LZWStream`, `AsciiHexStream`, `Ascii85Stream`, `RunLengthStream`) work without modification. The 100 MB default is large enough for any legitimate PDF content.

Fixes #1777. Replaces #1779 (which was auto-closed when the source fork was deleted).


Sharcoux:
Please bring this to [@cantoo/pdf-lib](https://www.npmjs.com/package/@cantoo/pdf-lib) for potential merge. It's the most advanced and maintained fork. This project is dead.

PR #1781, opened 2026-04-14
Fix integer overflow in XRef stream byte width parsing

### LINKED ISSUE #1776: Integer overflow in XRef stream byte width parsing leads to memory corruption
## Summary

The XRef stream parser in `src/core/parser/PDFXRefStreamParser.ts` reads byte width values from the `/W` array of a cross-reference stream dictionary without validation. These widths directly control loop iteration counts in `parseEntries()` (lines 84-127), allowing a crafted PDF to trigger integer overflow, excessive memory reads, and NaN propagation through parsed object references.

## Vulnerability Details

**CWE-190: Integer Overflow or Wraparound**

In the constructor (line 54-58), the `/W` array values are read as-is:

```typescript
const W = this.dict.lookup(PDFName.of('W'), PDFArray);
this.byteWidths = [-1, -1, -1];
for (let idx = 0, len = W.size(); idx < len; idx++) {
  this.byteWidths[idx] = W.lookup(idx, PDFNumber).asNumber();
}
```

These unvalidated widths are then used as loop bounds in `parseEntries()` (lines 96-109):

```typescript
let type = 0;
for (let idx = 0, len = typeFieldWidth; idx < len; idx++) {
  type = (type << 8) | this.bytes.next();
}

let offset = 0;
for (let idx = 0, len = offsetFieldWidth; idx < len; idx++) {
  offset = (offset << 8) | this.bytes.next();
}

let generationNumber = 0;
for (let idx = 0, len = genFieldWidth; idx < len; idx++) {
  generationNumber = (generationNumber << 8) | this.bytes.next();
}
```

### Attack scenario

A malicious PDF with a crafted `/W` array such as `[100, 100, 100]` causes:

1. **Excessive byte reads**: Each entry reads `typeFieldWidth + offsetFieldWidth + genFieldWidth` bytes from the stream. With large widths, ByteStream reads far past the actual stream data.
2. **Integer overflow via bit shifting**: JavaScript's `<<` operator works on 32-bit integers. After 4 iterations of `(type << 8)`, the value overflows 32-bit signed integer range. After ~8 iterations, values become inconsistent. With widths > 32, the left shift `<< 8` produces 0, and the accumulated value becomes just the last `| this.bytes.next()`.
3. **NaN propagation**: When `this.bytes.next()` reads past the end of the underlying buffer, it returns `undefined`. The expression `(offset << 8) | undefined` evaluates to `NaN` in some paths, which propagates into `PDFRef.of(objectNumber, generationNumber)` creating corrupted references.

### Impact

- Denial of service via excessive CPU and memory consumption during `PDFDocument.load()`
- Corrupted internal object table with NaN-based references, leading to unpredictable behavior
- Potential for crafted object references to alias unintended PDF objects

### Suggested fix

Validate byte widths after reading from the `/W` array:

```typescript
const W = this.dict.lookup(PDFName.of('W'), PDFArray);
this.byteWidths = [-1, -1, -1];
for (let idx = 0, len = W.size(); idx < len; idx++) {
  const width = W.lookup(idx, PDFNumber).asNumber();
  if (width < 0 || width > 8) {
    throw new Error(`Invalid XRef stream byte width: ${width} (must be 0-8)`);
  }
  this.byteWidths[idx] = width;
}
```

The PDF specification (ISO 32000-1:2008, Table 17) states that byte widths represent the number of bytes for each field in the cross-reference entry. Values above 8 bytes exceed the precision of JavaScript's number type and are not meaningful.

## Version

Tested against pdf-lib 1.17.1 (latest).

## Environment

Node.js

eddieran:
## Summary

Fixes #1776 -- integer overflow in XRef stream byte width parsing.

The `/W` array values from XRef streams are read without validation. These widths control loop bounds in `parseEntries()`. A crafted PDF with `/W [100, 100, 100]` causes excessive byte reads, integer overflow via bit shifting, and NaN propagation.

## Changes

- **Validate `/W` byte widths** after reading from the array. Each width must be 0--8 per the PDF spec (ISO 32000-1:2008, Table 17). Throws `InvalidXRefStreamByteWidthError` for out-of-range values.
- **Bounds-check `bytes.next()` in `parseEntries()`** to gracefully handle premature end-of-stream instead of propagating `undefined` through bitwise operations.
- **Added tests** for both the rejection of invalid widths and acceptance of valid widths.

## Files changed

- `src/core/errors.ts` -- new `InvalidXRefStreamByteWidthError` class
- `src/core/parser/PDFXRefStreamParser.ts` -- validation + bounds checking
- `tests/core/parser/PDFXRefStreamParser.spec.ts` -- 2 new test cases

## Test plan

- [x] All existing XRef stream parser tests pass (7/7)
- [x] New test: `/W [100, 100, 100]` throws descriptive error
- [x] New test: `/W [1, 4, 0]` (valid widths) does not throw


Sharcoux:
Please bring this to [@cantoo/pdf-lib](https://www.npmjs.com/package/@cantoo/pdf-lib) for potential merge. It's the most advanced and maintained fork. This project is dead.

